PR #2, opened 2018-03-07
Heroku

zingales:
Added some boilerplate code to get pushing to Heroku working. Also modified the port it was listening on, because it seems like Heroku defines the port for you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,4 @@ Temporary Items
 .apdisk
 
 
-# End of https://www.gitignore.io/api/go,osxCerealNotes
-
-CerealNotes
+# End of https://www.gitignore.io/api/go,osx
--- a/.gitignore
+++ b/.gitignore
@@ -45,4 +45,6 @@ Temporary Items
 .apdisk
 
 
-# End of https://www.gitignore.io/api/go,osx
+# End of https://www.gitignore.io/api/go,osxCerealNotes
+
+CerealNotes
--- a/index.go
+++ b/index.go
@@ -6,6 +6,14 @@ import (
     "net/http"
 )
 
+func determineListenAddress() (string, error) {
+  port := os.Getenv("PORT")
+  if port == "" {
+    return "", fmt.Errorf("$PORT not set")
+  }
+  return ":" + port, nil
+}
+
 func rootHandler(
         responseWriter http.ResponseWriter,
         request *http.Request) {
@@ -20,9 +28,17 @@ func rootHandler(
 }
 
 func main() {
+    addr, err := determineListenAddress()
+    if err != nil {
+        log.Fatal(err)
+    }
+    
     // set handler
     http.HandleFunc("/", rootHandler)
 
     // start server
-    log.Fatal(http.ListenAndServe(":80", nil))
+    log.Printf("Listening on %s...\n", addr)
+    if err := http.ListenAndServe(addr, nil); err != nil {
+        panic(err)
+    }
 }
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -45,4 +45,6 @@ Temporary Items
 .apdisk
 
 
-# End of https://www.gitignore.io/api/go,osx
\ No newline at end of file
+# End of https://www.gitignore.io/api/go,osxCerealNotes
```
atmiguel: why is `CerealNotes` at the end of this line?

.gitignore
```diff
@@ -45,4 +45,6 @@ Temporary Items
...
+
+CerealNotes
```
atmiguel: Is this popping up in the home directory? And not your bin directory?

index.go
```diff
@@ -4,8 +4,25 @@ import (
     "fmt"
     "log"
     "net/http"
+    "os"
+    "strconv"
 )
 
+// get the current listening address or fail if input is not correct
+func determineListenAddress() (string, error) {
+  portStr := os.Getenv("PORT")
```
atmiguel: let's do 4 spaces instead of 2

index.go
```diff
@@ -4,8 +4,25 @@ import (
...
+  
+  if portStr == "" {
+    return "", fmt.Errorf("$PORT not set")
+  }
+  
+  port, err := strconv.Atoi(portStr)
+  if err == nil {
+    return "", fmt.Errorf("$PORT not a valid integer")
```
atmiguel: We could also verify that the Port value is in the range 1024 – 49151, but seems unnecessary

index.go
```diff
@@ -20,9 +37,18 @@ func rootHandler(
 }
 
 func main() {
+    addr, err := determineListenAddress()
```
atmiguel: why not `port` instead of `addr`?

index.go
```diff
@@ -20,9 +37,18 @@ func rootHandler(
...
+        log.Fatal(err)
+    }
+
     // set handler
     http.HandleFunc("/", rootHandler)
 
     // start server
-    log.Fatal(http.ListenAndServe(":8080", nil))
+    log.Printf("Listening on %s...\n", addr)
+    err := http.ListenAndServe(addr, nil)
```
atmiguel: I think this was better when it was part of the if statement

atmiguel (changes requested):

.gitignore
zingales: Ya its cause i did go build install of go install. We want to use go install for sure. I'm not sure i understand why its like that, but i don't have a strong feeling about it

index.go
zingales: ya honestly i felt a bit overboard even validating that it was an int. Maybe i should remove the code. I just feel weird with this arbitrary value set in the env.

index.go
zingales: its an address cause it has a preceding ":" and its a string

index.go
zingales: will change to if block as discussed.

.gitignore
atmiguel: so we planning on removing this?

index.go
atmiguel: why does it make you feel weird?

index.go
atmiguel: after some googling looks like something like `:80` is referred to as the port, and `80` is the port number.

.gitignore
zingales: yes

index.go
zingales: OhhHhHhhHh will change method and variables to port then

atmiguel (approved):

PR #3, opened 2018-03-08
Added html templates

atmiguel:


README.md
```diff
@@ -4,6 +4,8 @@ Assuming your local environment is setup correctly with Golang standards, you ca
 
 `cd to this repo`
 
+`export PORT=8080`
```
zingales, line 7: So Heorku uses environment variables for configurations. And it also has a deploy local command, that will populate certain environment fields from a json file. I'll look into it but once i've set that up we will no longer need to do this manually.

main.go
```diff
@@ -10,25 +11,21 @@ import (
...
 
-    var responseContent string = fmt.Sprintf(
-        "Hi there, I love %s!",
-        request.URL.Path[1:])
-
-    fmt.Fprintf(
-        responseWriter,
-        responseContent)
+    // TODO handle error
+    parsedTemplate, _ := template.ParseFiles("root.tmpl")
```
zingales: I see you have the todo, but can we default to just log errors? I just don't like silent bugs, i'd rather have to sift through too many logs than have to find an error that is being swallowed

root.tmpl
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
zingales, line 1: Lets make a separate folder for templates.

zingales (changes requested):

root.tmpl
atmiguel: done in a future branch

main.go
atmiguel: will do. though how do we decide when to `log.Fatal()` vs `panic()`?

zingales (approved):

main.go
zingales: I think panic should be avoided unless the program literally cannot continue. 

https://www.reddit.com/r/golang/comments/2a1sie/when_to_call_panic/ciqn81u/

atmiguel: okay we've gotta figure out a succinct way of returning a 500 in these situations

PR #4, opened 2018-03-08
Add static files

atmiguel:


main.go
```diff
@@ -19,27 +20,68 @@ func determineListenPort() (string, error) {
...
+                responseWriter,
+                []string{http.MethodGet})
+    }
 }
 
 func main() {
+    // SET ROUTER
+
+    // static files
+    {
```
zingales, line 63: This block is just for handling static assets? So anything we request that is static will automatically be redirected and go will look for things in `./static/` folder? maybe add a comment for the go uninitiated like me

templates/login_or_signup.tmpl
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+
+<html lang="en">
+    <head>
+        <meta charset="utf-8" />
+        <meta
+            name="description"
+            content="The best way to book club" />
+        <meta
```
zingales, line 9: just curious what does this do?

zingales (approved):

main.go
atmiguel: I wouldn't say "redirected", but yeah your understanding is correct. This was the standard way to treat static files in Django as well. What comment would have helped you understand this code quicker?

templates/login_or_signup.tmpl
atmiguel: Yeah so I had to look this boilerplate shit up. I can talk about every line honestly, if you're curious about the other meta tags as well. This one sets the text that shows up under our site when you search for it on google or whatever. You can also see the description in the fb link popup
screen shot 2018-03-08 at 1 17 44 pm

templates/login_or_signup.tmpl
```diff
@@ -0,0 +1,21 @@
...
+            name="viewport"
+            content="width=device-width, initial-scale=1" />
```
zingales, line 11: what about this one?

main.go
zingales: its probably cause i'm not used to it, but static file handler boilerplate is what i would have put it.

templates/login_or_signup.tmpl
atmiguel: I understand this one less, but it has something to do with setting the width of the page. See: https://css-tricks.com/snippets/html/responsive-meta-tag/

PR #6, opened 2018-03-09
Connect to database

zingales:
Added boilerplate code to connect to db in Go and tested with a simple ping. 

Added code facilitate running heroku local. 

Added required vendor dir to git.


.gitignore
```diff
@@ -45,4 +45,7 @@ Temporary Items
 .apdisk
 
 
-# End of https://www.gitignore.io/api/go,osx
\ No newline at end of file
+# End of https://www.gitignore.io/api/go,osx
+
+### GoDep ###
+GoDeps/Readme
```
atmiguel, line 51: what's in the godeps readme?

.env
```diff
@@ -0,0 +1,2 @@
+DATABASE_URL='localhost'
```
atmiguel, line 1: how does this file work?

Godeps/Godeps.json
```diff
@@ -2,5 +2,16 @@
 	"ImportPath": "CerealNotes",
 	"GoVersion": "go1.10",
 	"GodepVersion": "v80",
-	"Deps": []
+	"Deps": [
+		{
```
atmiguel, line 6: any way we could do spaces instead of tabs here?

cerealNotesDb/cerealNotesDb.go
```diff
@@ -0,0 +1,22 @@
+package cerealNotesDb
```
atmiguel: apparently we're not supposed to use camelCase package names: https://blog.golang.org/package-names

main.go
```diff
@@ -127,6 +128,9 @@ func main() {
 				fileServer))
 	}
 
+	db := cerealNotesDb.Connect(os.Getenv("DATABASE_URL"))
+	db.Ping()
```
atmiguel: did you mean to catch the error here or something?

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE IF NOT EXISTS "Users"(
```
atmiguel: why use `IF NOT EXISTS`?

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
...
+	id bigserial PRIMARY KEY,
```
atmiguel: how strongly do you feel about these all caps keywords? i would likely prefer to just have all lowercase

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
...
+    password bytea NOT NULL,
+    "creationTime" timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS "Publications"(
+	id bigserial PRIMARY KEY,
+    "authorId" bigint references "Users"(id) NOT NULL,
+    "sequenceNumber" serial NOT NULL,
+	"creationTime" timestamp NOT NULL
+
```
atmiguel: extra line

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE IF NOT EXISTS "Users"(
+	id bigserial PRIMARY KEY,
+    "displayName" text NOT NULL,
+    "emailAddress" text UNIQUE NOT NULL,
+    password bytea NOT NULL,
+    "creationTime" timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS "Publications"(
+	id bigserial PRIMARY KEY,
```
atmiguel: looks like you're mixing tabs and spaces in here. things aren't lining up

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
...
+    "authorId" bigint references "Users"(id) NOT NULL,
```
atmiguel: why quotes around Users?

atmiguel: omg is there some bullshit about postgres ignoring your case unless it's in quotes? if that's the case, i say fuck all capitalization in this file

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
...
+    "sequenceNumber" serial NOT NULL,
+	"creationTime" timestamp NOT NULL
+
+);
+
+CREATE TABLE IF NOT EXISTS "Notes"(
+	id bigserial PRIMARY KEY,
```
atmiguel: tab

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
...
+    "authorId" bigint references "Users"(id) NOT NULL,
+    "type" text,
```
atmiguel: enum?

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,24 @@
...
+    "displayName" text NOT NULL,
+    "emailAddress" text UNIQUE NOT NULL,
+    password bytea NOT NULL,
+    "creationTime" timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS "Publications"(
+	id bigserial PRIMARY KEY,
+    "authorId" bigint references "Users"(id) NOT NULL,
+    "sequenceNumber" serial NOT NULL,
```
atmiguel: I was thinking about this again. I don't think we need this field. The query is pretty straightforward without it:
```
select * from publications
    where author_id = _user_id
    order by id asc
    limit _number_of_publications_to_show
```

atmiguel (changes requested):
Did you mean to add those lib/pq files? Should they be a level or two up in the file hierarchy?

migrations/0000_createDbs.sql
zingales: If not exists is so that it's idempotent. So you can rerun this script over and over without any issues.

migrations/0000_createDbs.sql
zingales: I thought SQL was supposed to have all caps keywords. But i don't particularly care

migrations/0000_createDbs.sql
zingales: yes there is bullshit about caps without quotes. Want me to make all of the tables + fields lower case? so like authoerid?

migrations/0000_createDbs.sql
zingales: Didn't know enum was a thing in postgres will update to include that.

.env
zingales: when you run heroku local, it basically uses the .env file to temporarily initialize the environment variables such that os.get_en("PORT") works

cerealNotesDb/cerealNotesDb.go
zingales: Okay will update, what do you think of the name? maybe just database?

main.go
zingales: I should probably remove this. I already do the ping in the actual connect.

.gitignore
zingales: It literally says do not change, this file is generated automatically, if you want your very own i think running godeps save will generate it.

cerealNotesDb/cerealNotesDb.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"database/sql"
+	// Notice that we’re loading the driver anonymously, The driver registers itself as being available to the database/sql package.
+	_ "github.com/lib/pq"
+	"log"
+)
+
+func Connect(dbUrl string) *sql.DB {
+	db, err := sql.Open("postgres", dbUrl)
+	if err != nil {
```
zingales: will change this to return db, err instead of just calling log.fatal.

.env
atmiguel: what do you mean when you run heroku local? is `heroku local` a command?

Godeps/Godeps.json
atmiguel: nevermind, didn't realize this was auto-generated

migrations/0000_createDbs.sql
atmiguel: oh interesting. did you read something saying you should make your sql files idempotent?

migrations/0000_createDbs.sql
atmiguel: yeah nevermind, the internet is saying to do what you did. it just seems unnecessarily annoying, but i guess it's convention

migrations/0000_createDbs.sql
atmiguel: yeah looks like the convention is underscores so let's do `users` and `author_id`

.env
zingales: yes `heroku local` runs the commands in the Procfile uses the environment variables in .env

main.go
zingales: I added it cause otherwise go complains that db is not being used.

.env
atmiguel: i'm sorry, so just so i fully understand, when do you run `heroku local`? do i ever need to do that on my machine? does the heroku instance somehow know to run that when it's starting up?

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,27 @@
+-- Types
+CREATE TYPE notetype AS ENUM ('Predictions', 'Marginalia', 'Meta', 'Questions');
```
atmiguel: how about `note_type`? also, i think we should lowercase the enum values

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,27 @@
...
+	email_address text UNIQUE NOT NULL,
+	password bytea NOT NULL,
+	creation_time timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS publications(
+	id bigserial PRIMARY KEY,
+	author_id bigint references users(id) NOT NULL,
+	creationTime timestamp NOT NULL
+
```
atmiguel: remove this newline

migrations/0000_createDbs.sql
```diff
@@ -0,0 +1,27 @@
...
+	display_name text NOT NULL,
+	email_address text UNIQUE NOT NULL,
+	password bytea NOT NULL,
+	creation_time timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS publications(
+	id bigserial PRIMARY KEY,
+	author_id bigint references users(id) NOT NULL,
+	creationTime timestamp NOT NULL
```
atmiguel: forgot to underscore, instead of camelCase

atmiguel (changes requested):

.env
zingales: Its cool. You don't need to run it. But when i test on my machine I run `heroku local`. If you have everything set up correctly it just starts the our web server.

atmiguel (approved):

PR #11, opened 2018-03-18
Separated out postgres functions

atmiguel:
I don't know the best way to organize the code. Imma read up on MVC and figure out what makes sense

### DIFF
--- a/databaseutil/databaseutil.go
+++ b/databaseutil/databaseutil.go
@@ -54,7 +54,6 @@ func InsertIntoUsersTable(
 
 	var userId user.UserId
 	if err := row.Scan(&userId); err != nil {
-		// TODO handle err.ErrNoRows
 		return -1, err
 	}
 
@@ -73,7 +72,6 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
 
 	var password []byte
 	if err := row.Scan(&password); err != nil {
-		// TODO handle err.ErrNoRows
 		return nil, err
 	}
 
--- a/databaseutil/databaseutil.go
+++ b/databaseutil/databaseutil.go
@@ -2,30 +2,17 @@ package databaseutil
 
 import (
 	"database/sql"
-	"github.com/atmiguel/cerealnotes/models/user"
 	_ "github.com/lib/pq"
-	"log"
-	"os"
 	"time"
 )
 
 var db *sql.DB
 
-func init() {
-	var databaseUrl string
-	{
-		environmentVariableName := "DATABASE_URL"
-		databaseUrl = os.Getenv(environmentVariableName)
-
-		if len(databaseUrl) == 0 {
-			log.Fatalf("environment variable %s not set", environmentVariableName)
-		}
-	}
-
+func ConnectToDatabase(databaseUrl string) error {
 	{
 		tempDb, err := sql.Open("postgres", databaseUrl)
 		if err != nil {
-			log.Fatal(err)
+			return err
 		}
 
 		db = tempDb
@@ -33,8 +20,10 @@ func init() {
 
 	// Quickly test if the connection to the database worked.
 	if err := db.Ping(); err != nil {
-		log.Fatal(err)
+		return err
 	}
+
+	return nil
 }
 
 func InsertIntoUsersTable(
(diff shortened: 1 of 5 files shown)


databaseutil/databaseutil.go
```diff
@@ -2,82 +2,78 @@ package databaseutil
...
 	"log"
+	"os"
 	"time"
 )
 
 var db *sql.DB
 
-func Connect(dbUrl string) error {
-	var err error
+func init() {
```
zingales: I'm not the biggest fan of the init doing all this db connection. I also except this will make testing more difficult in case you ever want to pass in a mock database. But I agree the way we had it before with connect was also bad.

databaseutil/databaseutil.go
```diff
@@ -2,82 +2,78 @@ package databaseutil
...
-	err = db.QueryRow(
-		sqlQuery,
-		displayName,
-		emailAddress,
-		hashedPassword,
-		time.Now().UTC(),
-	).Scan(&id)
-	if err != nil {
+	var userId user.UserId
+	if err := row.Scan(&userId); err != nil {
```
zingales: Thanks for fixing this issue!

main.go
```diff
@@ -88,20 +84,15 @@ func handleUserRequest(
 			panic(err)
 		}
 
-		userId, err := databaseutil.CreateUser(
+		_, err := userservice.CreateUser(
```
zingales: this line highlights that create user is improperly named. It should be called add user to database, or it should create the user as an object return it then have a second line that adds it to the database. Here esp with the anonymous var it looks like we are specifically throwing away the newly created user.

services/userservice/userservice.go
```diff
@@ -0,0 +1,46 @@
+package userservice
```
zingales, line 1: not a huge fan of service, but can't think of anything better. Anything else you were considering?

zingales (changes requested):

databaseutil/databaseutil.go
atmiguel: fair enough. any suggestions for what to do instead? perhaps while trying to write tests for this i'll discover something that makes sense here?

main.go
atmiguel: hmm... I think we have very different mental models for the code. In my mind, a user is something that only truly exists in the database. So saying "create" user to me means that it is being inserted into the database. I would think it very misleading if `createUser` just constructed an object in the code.

I'm also just very used to the idea that `createX` inserts it into the database and returns the id. Would it be clearer if it was `recordX` or something?

Another thing that may be causing confusion here is that we're conflating two things in the user table. We could theoretically have a user table that just has displayName (and perhaps emailAddress), and then an authentication or password table with userId and password

services/userservice/userservice.go
atmiguel: i'm honestly not sure if i like the way the codebase is structured. In the next branch I also called things handlers instead of controllers (as the MVC pattern says we should do).

I'm open to suggestions. I'm not a huge fan of `userutil`. Just `user` might be okay, but it's in use by the model package.

databaseutil/databaseutil.go
zingales: I suspect creating our own db wrapper object might make sense. So you call connect and it returns a "DBConnectionHelper" or something like that. Then we change all the functions to methods of that type/class. Then in the code have main.go initialize a instance config which gets passed around. 

 Another crazy idea is having the config object initialized in main.go that stores function pointers, instead of like static info. So you can do like config.GetConnection() and it will return a db connection. That way its super easy to create mock types, because you'll just pass a new config object, with constructors for whatever type you want. .

main.go
zingales: Hmmm, ya i think i'm used to `create` in the more of an OOP way. So like `createX` would be a static method on a class that would return a new object of that class. Also to exacerbate when dealing with db i'm used to transactions so you'd have to call save at the end of your operations for the information to be stored into the db. I highly prefer `recordX` over `createX`.  What are your feelings on `saveX` or `storeX`.

services/userservice/userservice.go
zingales: what if we just put all these functions into the user package? 
or potentially just put it into the `usercontroller` file, does anything other than the user controller need these functions? 

Other not great list of names
`userwrapper`, `usermanager`, `userhelper`

databaseutil/databaseutil.go
atmiguel: Remove init and recreate Connect function

main.go
atmiguel: Change to `storeX` and no longer return id

atmiguel: `storeNewX`

databaseutil/databaseutil.go
atmiguel: `authenticateUserCredentials`

zingales (approved):

PR #12, opened 2018-03-18
Separate out router and handlers

atmiguel:


zingales (approved):
It looks like its all just reorganizing functions. If that's the case approved!

PR #13, opened 2018-03-24
Alert user when email already in use

atmiguel:


services/userservice/userservice.go
```diff
@@ -1,11 +1,14 @@
 package userservice
 
 import (
+	"errors"
 	"github.com/atmiguel/cerealnotes/databaseutil"
 	"golang.org/x/crypto/bcrypt"
 	"time"
 )
 
+var EmailAddressAlreadyInUseError = errors.New("")
```
zingales: Why not put an error message here instead of empty string?

databaseutil/databaseutil.go
```diff
@@ -46,6 +49,12 @@ func InsertIntoUsersTable(
 			return nil
 		}
 
+		if pgErr, ok := err.(*pq.Error); ok {
+			if pgErr.Code == "23505" {
+				return UniqueConstraintError
```
zingales: Considering that in  [create new user](https://github.com/atmiguel/cerealnotes/pull/13/files#diff-09499008c3b5f7673ff66923b8758afaR32) we convert this error (regardless of which unique constraint is violated) into a EmailAddressAlreadyInUseError.  Why not throw  EmailAddressAlreadyInUseError here instead.

databaseutil/databaseutil.go
```diff
@@ -46,6 +49,12 @@ func InsertIntoUsersTable(
 			return nil
 		}
 
+		if pgErr, ok := err.(*pq.Error); ok {
+			if pgErr.Code == "23505" {
```
zingales: Consider making this a const int at the top of the file.

static/js/login_or_signup.js
```diff
@@ -125,7 +125,13 @@ $(function() {
             if (request.status === 201) {
                 alert('Successfully created user');
             } else {
+                alert('Unknown successful status');
```
zingales: Is all this alerting temporary? If so consider making a task to take care of it.

zingales (changes requested):

services/userservice/userservice.go
zingales: Also in my PR i added https://github.com/pkg/errors. With this Library creating errors is easier. But I think you'll need to create a function that returns a new error. Otherwise the stacktrace for this error will always be line 10 in userservice

atmiguel: I suppose I was just thinking that the error message should never be displayed, because this error should always be checked for, but that's not a great reason. I'll add a message.

I'm not clear on what pkg/errors does. Looks like it makes it easier to give context to custom errors? We don't need context here.

Also, I'm not sure what you mean by creating a function that returns a new error?

databaseutil/databaseutil.go
atmiguel: That link doesn't go anywhere for me btw.

`UniqueConstraintError` is a database level error that I expect will be thrown by future functions. I'm expecting that we'll abstract away a lot of postgres errors, because looking up the specific error codes every time is a recipe for disaster

databaseutil/databaseutil.go
atmiguel: it's a string. See above comment about plans for this code. I was planning on waiting for another use case before generalizing though

static/js/login_or_signup.js
atmiguel: I mean, all the alerts are temporary. But yeah I'm not sure what the right thing to do is here. Perhaps we need to create a general "something went wrong with that action" popup or something? This case is especially weird, because it would mean the server returned a 2XX code, but one that was unexpected. Like wtf.

services/userservice/userservice.go
atmiguel: I suppose I could add which email address it is to the context. Though again, the expectation is that the error is caught immediately

static/js/login_or_signup.js
zingales: Ya too much for this code review but maybe create a task to add cool little banner notifications that fade on certain actions. Something kinda like [this](https://css-tricks.com/pop-from-top-notification/)

databaseutil/databaseutil.go
zingales: fixed the link. I guess it just feels weird throwing an error that we always cast immediately to another error type.

services/userservice/userservice.go
```diff
@@ -26,6 +29,10 @@ func StoreNewUser(
 		hashedPassword,
 		creationTime,
 	); err != nil {
+		if err == databaseutil.UniqueConstraintError {
```
zingales, line 33: Also this would be a good use of [error.Wrap](https://godoc.org/github.com/pkg/errors#Wrap) <-- this is a new package btw, but i already included it in my PR.  Also i suspect that your throwing a singular error around, instead of creating a new one each time this situation happens happens. This means, on top of other issues,  that the stack trace here is not going to make sense.

databaseutil/databaseutil.go
zingales: Makes sense though i'd still prefer it to be a const string.

static/js/login_or_signup.js
atmiguel: created https://trello.com/c/8VCLI26b

services/userservice/userservice.go
atmiguel: That link doesn't work for me

zingales: this is the link i meant: https://godoc.org/github.com/pkg/errors#Wrap

services/userservice/userservice.go
zingales: I did some testing and we can ignore the function part, since normal go errors don't have any stacktrace-like information. 

the package however, bascially allows you to slowly build a stack trace by using the wrap command. Which keeps a stack of information of what's been wrapped . So we can eventually see that this error was caused by a different in a function we called and then wrapped into another error .

services/userservice/userservice.go
zingales: On second thought this probably deserves a full conversation on its own, and blocking this PR for this issues doesn't make sense. I've made a task to talk about it futher: https://trello.com/c/oR4o9rh1/35-discussion-how-do-we-want-to-handler-error-handling

zingales (approved):

PR #14, opened 2018-03-25
Jwt Authentication

zingales:
Create, save, validate JWT on login.


README.md
```diff
@@ -1,22 +1,25 @@
...
-# Heroku stuffs
-
-* install godep: `go get github.com/tools/godep`
-* install heroku cli tools: `brew install heroku`
-
-# Db stuffs
-
-* run first migration `heroku pg:psql < migrations/0000_createDbs.sql`
+# Run DB migrations
+More db information in `migrations/ReadMe.md`
```
atmiguel: nbd, but should name the file `README.md`

databaseutil/databaseutil.go
```diff
@@ -69,3 +69,21 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
 
 	return password, nil
 }
+
+func GetUserIdFromUserWithEmailAddress(emailAddress string) (int64, error) {
```
atmiguel: We should return `UserId` type

databaseutil/databaseutil.go
```diff
@@ -69,3 +69,21 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
...
+		sqlQuery := `
+			SELECT id FROM users
+			WHERE email_address = $1`
+
+		row = db.QueryRow(sqlQuery, emailAddress)
+	}
+
+	var userId int64
+	if err := row.Scan(&userId); err != nil {
+		return 0, err
```
atmiguel: i think returning -1 makes more sense.

handlers/handlers.go
```diff
@@ -3,21 +3,43 @@ package handlers
...
 	"html/template"
 	"io/ioutil"
 	"log"
 	"net/http"
 	"strings"
+	"time"
 )
 
+type CerealNotesClaims struct {
+	UserId models.UserId `json:"UserId"`
```
atmiguel: Pretty sure you can remove the first `UserId` from this line. The name is taken from the type if there is no name.

atmiguel: so far our convention for json is camelCase

handlers/handlers.go
```diff
@@ -3,21 +3,43 @@ package handlers
...
+	"github.com/pkg/errors"
 	"html/template"
 	"io/ioutil"
 	"log"
 	"net/http"
 	"strings"
+	"time"
 )
 
+type CerealNotesClaims struct {
```
atmiguel: Can we try to think of another name for this? I remember naming things `BevSpotX` and I feel like we would just fall back on it when we couldn't think of a better name

handlers/handlers.go
```diff
@@ -3,21 +3,43 @@ package handlers
...
+	UserId models.UserId `json:"UserId"`
+	jwt.StandardClaims
+}
+
+var tokenSigningKey []byte
+
+func SetSigningKey(key []byte) {
```
atmiguel: not `SetTokenSigningKey`?

handlers/handlers.go
```diff
@@ -3,21 +3,43 @@ package handlers
...
 // HANDLERS
 func HandleLoginOrSignupRequest(
 	responseWriter http.ResponseWriter,
 	request *http.Request,
 ) {
 	switch request.Method {
 	case http.MethodGet:
+		// Check to see if they are already logged in if so redirect
+		if _, err := getUserIdFromStoredToken(request); err == nil {
+			http.Redirect(responseWriter, request, paths.HomePath, http.StatusSeeOther)
```
atmiguel: idt SeeOther is the right code for this. I'm not sure which one is correct though: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+			if err != nil {
+				panic(err)
+			}
+
+			token, err := createTokenAsString(userId, oneWeekInMinutes)
+			if err != nil {
+				panic(err)
+			}
+
+			expiration := time.Now().Add(oneWeekInMinutes * time.Minute)
```
atmiguel: this is weird. What is `time.Minute`?

atmiguel: How about `expirationTime`?

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+
+			cookie := http.Cookie{
+				Name:     cerealNotesCookieName,
+				Value:    token,
+				Expires:  expiration,
+				HttpOnly: true,
+			}
+
+			http.SetCookie(responseWriter, &cookie)
+
```
atmiguel: remove empty line

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+				panic(err)
+			}
+
+			expiration := time.Now().Add(oneWeekInMinutes * time.Minute)
+
+			cookie := http.Cookie{
+				Name:     cerealNotesCookieName,
+				Value:    token,
+				Expires:  expiration,
+				HttpOnly: true,
```
atmiguel: what's this?

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+
+			http.SetCookie(responseWriter, &cookie)
+
+		}
+
 		responseWriter.WriteHeader(http.StatusCreated)
 		responseWriter.Write([]byte(fmt.Sprint("passward email combo was correct")))
 
+	case http.MethodDelete:
+		//Cookie will overwrite existing cookie then delete itself
```
atmiguel: what's this? since jwt is stateless, the logout action doesn't even need to hit the backend right?

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+				Value:    "",
+				HttpOnly: true,
+				MaxAge:   -1,
+			}
+
+		http.SetCookie(responseWriter, &cookie)
+		responseWriter.WriteHeader(http.StatusOK)
+		responseWriter.Write([]byte(fmt.Sprint("user succefully logged out")))
 	default:
 		respondWithMethodNotAllowed(responseWriter, []string{http.MethodPost})
```
atmiguel: `http.MethodDelete` should be added here

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
 	}
 }
 
+type AuthentictedRequestHandlerType func(http.ResponseWriter, *http.Request, models.UserId)
+
+func AuthenticateOrRedirectToLogin(
+	authenticatedHandlerFunc AuthentictedRequestHandlerType,
+) func(http.ResponseWriter, *http.Request) {
+	return http.HandlerFunc(
```
atmiguel: why do we need to call `HandlerFunc`?

handlers/handlers.go
```diff
@@ -3,21 +3,43 @@ package handlers
...
+}
+
 // HANDLERS
 func HandleLoginOrSignupRequest(
 	responseWriter http.ResponseWriter,
 	request *http.Request,
 ) {
 	switch request.Method {
 	case http.MethodGet:
+		// Check to see if they are already logged in if so redirect
```
atmiguel: let's add punctuation to our comments

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
 			panic(err)
 		}
 
+		// Create token as a cookie and set it
```
atmiguel: token as a cookie?

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
 
+type AuthentictedRequestHandlerType func(http.ResponseWriter, *http.Request, models.UserId)
+
+func AuthenticateOrRedirectToLogin(
+	authenticatedHandlerFunc AuthentictedRequestHandlerType,
+) func(http.ResponseWriter, *http.Request) {
+	return http.HandlerFunc(
+		func(responseWriter http.ResponseWriter, request *http.Request) {
+			if userId, err := getUserIdFromStoredToken(request); err == nil {
+				// call authenticatedHandlerFunc if log in works
```
atmiguel: login should be one word

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+				authenticatedHandlerFunc(responseWriter, request, userId)
+			} else {
+				// Else redirect to login page
+				http.Redirect(
+					responseWriter,
+					request,
+					paths.LoginSignupPath,
+					http.StatusSeeOther,
+				)
```
atmiguel: this goes up one line

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
 }
 
+type AuthentictedRequestHandlerType func(http.ResponseWriter, *http.Request, models.UserId)
+
+func AuthenticateOrRedirectToLogin(
+	authenticatedHandlerFunc AuthentictedRequestHandlerType,
+) func(http.ResponseWriter, *http.Request) {
+	return http.HandlerFunc(
+		func(responseWriter http.ResponseWriter, request *http.Request) {
+			if userId, err := getUserIdFromStoredToken(request); err == nil {
```
atmiguel: could do `err != nil` since it's typically what's expected, but it's nbd either way

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
 
 	return body
 }
+
+
```
atmiguel: remove empty line

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+// Token Util
+const oneWeekInMinutes = 60 * 24 * 7
```
atmiguel: these should go at the top of the file

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+const cerealNotesCookieName = "CerealNotesToken"
+
+func parseTokenFromString(tokenString string) (*jwt.Token, error) {
+	// Parse the token
```
atmiguel: no need for this comment

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+	token, err := jwt.ParseWithClaims(
```
atmiguel: could just return here

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+		strings.TrimSpace(tokenString),
+		&CerealNotesClaims{},
+		func(token *jwt.Token) (interface{}, error) {
```
atmiguel: how does golang not throw an error here? `token` is unused

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+	claims := CerealNotesClaims{
+		userId,
+		jwt.StandardClaims{
+			ExpiresAt: time.Now().Unix() + (expirationTimeInMinutes * 60),
+			Issuer:    "CerealNotes",
+		},
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	ss, err := token.SignedString(tokenSigningKey)
```
atmiguel: could just return this line

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+	return ss, err
+}
+
+func getUserIdFromStoredToken(request *http.Request) (models.UserId, error) {
+	cookie, err := request.Cookie(cerealNotesCookieName)
+	if err != nil {
+		return 0, err
```
atmiguel: -1 on all these returns i think

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+	}
+
+	token, err := parseTokenFromString(cookie.Value)
+	if err != nil {
+		return 0, err
+	}
+
+	if claims, ok := token.Claims.(*CerealNotesClaims); ok && token.Valid {
+		return claims.UserId, nil
+	} else {
```
atmiguel: no need for this else, since the if statement is returning

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+		return 0, errors.Errorf("Token was invalid or unreadable")
+	}
+}
+
+func tokenTest1() {
```
atmiguel: remove? or move to another file?

main.go
```diff
@@ -33,6 +37,19 @@ func determineDatabaseUrl() (string, error) {
 	return databaseUrl, nil
 }
 
+func determineSigningKey() ([]byte, error) {
+	signingKeyVariableName := "TOKEN_SIGNING_KEY"
```
atmiguel: why is it sometimes token signing key and sometimes just signing key?

main.go
```diff
@@ -48,6 +65,16 @@ func main() {
...
 	}
 
+	// set up signing key
+	{
+		tokenSigningKey, err := determineSigningKey()
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		handlers.SetSigningKey(tokenSigningKey)
```
atmiguel: reading this, it seems like `SetSigningKey` may not belong in `handlers`

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+
+		http.SetCookie(responseWriter, &cookie)
+		responseWriter.WriteHeader(http.StatusOK)
+		responseWriter.Write([]byte(fmt.Sprint("user succefully logged out")))
 	default:
 		respondWithMethodNotAllowed(responseWriter, []string{http.MethodPost})
 	}
 }
 
+type AuthentictedRequestHandlerType func(http.ResponseWriter, *http.Request, models.UserId)
```
atmiguel: this line is too long

paths/paths.go
```diff
@@ -0,0 +1,4 @@
+package paths
+
+const HomePath = "/home"
```
atmiguel, line 3: Could/Should these paths live in the routers package?

paths/paths.go
```diff
@@ -0,0 +1,4 @@
...
+const LoginSignupPath = "/login-or-signup"
```
atmiguel: Would prefer `LoginOrSignupPath`

routers/routers.go
```diff
@@ -19,9 +20,20 @@ func SetRoutes() {
...
-	http.HandleFunc("/login-or-signup", handlers.HandleLoginOrSignupRequest)
+	http.HandleFunc(paths.LoginSignupPath, handlers.HandleLoginOrSignupRequest)
 
 	// forms
 	http.HandleFunc("/user", handlers.HandleUserRequest)
 	http.HandleFunc("/session", handlers.HandleSessionRequest)
+
+	// requires Authentication
+	handleAuthenticated(paths.HomePath, handlers.HandleHomeRequest)
+
```
atmiguel: remove empty line

services/userservice/userservice.go
```diff
@@ -48,3 +49,12 @@ func AuthenticateUserCredentials(emailAddress string, password string) error {
 
 	return nil
 }
+
+func GetUserIdFromEmailAddress(emailAddress string) (models.UserId, error) {
```
atmiguel: Hmm... If you make the change I was recommending for the databaseutil function, then this function is basically moot. I'm not sure what makes sense here. I definitely feel weird about these two functions having such different names though. If we keep both of them, I think they should be called `GetIdForUserWithEmailAddress`

static/js/home.js
```diff
@@ -0,0 +1,13 @@
+
```
atmiguel: remove empty line

static/js/home.js
```diff
@@ -0,0 +1,13 @@
...
+$(function() {
+	$("#logout-button").click(() => {
```
atmiguel: can we use spaces instead of tabs?

static/js/login_or_signup.js
```diff
@@ -123,16 +123,24 @@ $(function() {
     attachSubmitClickHandler(signupFormMetadata, (formDataAsJsonString) => {
         $.post('/user', formDataAsJsonString, (responseBody, _, request) => {
             if (request.status === 201) {
-                alert('Successfully created user');
+                $("#signup-tab").removeClass("mui--is-active")
```
atmiguel: There's a more correct way of doing this. See here: https://www.muicss.com/docs/v1/css-js/tabs
Try: `mui.tabs.activate(paneId)`

templates/home.tmpl
```diff
@@ -0,0 +1,33 @@
...
+            content="width=device-width, initial-scale=1" />
+
+        <title>CerealNotes</title>
+
+        <link href="//cdn.muicss.com/mui-0.9.36/css/mui.min.css" rel="stylesheet" type="text/css" />
+
+        <script src="//cdn.muicss.com/mui-0.9.36/js/mui.min.js"></script>
+        <script src="//ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
+        <script src="/static/js/home.js"></script>
+    </head>
```
atmiguel, line 20: We'll deal with DRYing up this stuff in a later PR

README.md
```diff
@@ -1,22 +1,25 @@
...
+# Installation
+## Locally
+* postgres server installed and running. 
+	* `brew install postgres`
+	* `pg_ctl -D /usr/local/var/postgres start`
+* heroku cli installed
+	* `brew install heroku`
+* golang installed
+	* `brew install go`
+* godep installed: `go get github.com/tools/godep`
```
atmiguel: Stay consistent. Put the `go get` on the next line

README.md
```diff
@@ -1,22 +1,25 @@
-# CerealNotes
+# Installation
+## Locally
+* postgres server installed and running.
```
atmiguel: nbd. we can clean this up in a later PR, but this line ends with a period, the next two don't, and the fourth one ends with a colon

handlers/handlers.go
atmiguel: and say "user" in place of "they"

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
 		}
 
+		// Create token as a cookie and set it
+		{
+			userId, err := userservice.GetUserIdFromEmailAddress(loginForm.EmailAddress)
+			if err != nil {
+				panic(err)
+			}
+
+			token, err := createTokenAsString(userId, oneWeekInMinutes)
```
atmiguel: why do we need to specify one week in two places?

handlers/handlers.go
atmiguel: also, nitpick. non-code comments should have a space after the slashes

handlers/handlers.go
atmiguel: actually, is this comment necessary?

handlers/handlers.go
```diff
@@ -89,14 +111,92 @@ func HandleSessionRequest(
...
+func AuthenticateOrRedirectToLogin(
+	authenticatedHandlerFunc AuthentictedRequestHandlerType,
+) func(http.ResponseWriter, *http.Request) {
+	return http.HandlerFunc(
+		func(responseWriter http.ResponseWriter, request *http.Request) {
+			if userId, err := getUserIdFromStoredToken(request); err == nil {
+				// call authenticatedHandlerFunc if log in works
+				authenticatedHandlerFunc(responseWriter, request, userId)
+			} else {
+				// Else redirect to login page
```
atmiguel: seems pretty apparent

handlers/handlers.go
atmiguel: though we can discuss it, if you prefer this style

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
 
 	return body
 }
+
+
+// Token Util
+const oneWeekInMinutes = 60 * 24 * 7
+const cerealNotesCookieName = "CerealNotesToken"
+
+func parseTokenFromString(tokenString string) (*jwt.Token, error) {
```
atmiguel: is it `tokenAsString` or `tokenString`? Let's be consistent

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+			return tokenSigningKey, nil
+		})
+	return token, err
+}
+
+func createTokenAsString(
+	userId models.UserId,
+	expirationTimeInMinutes int64,
+) (string, error) {
+	// Create the Claims
```
atmiguel: unnecessary comment

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+	claims := CerealNotesClaims{
+		userId,
+		jwt.StandardClaims{
+			ExpiresAt: time.Now().Unix() + (expirationTimeInMinutes * 60),
```
atmiguel: wait so the cookie has the expiration time as a timestamp, and in the token it's seconds since the epoch? I know we don't get to decide those things, but jesus

handlers/handlers.go
```diff
@@ -121,3 +221,79 @@ func getRequestBody(request *http.Request) []byte {
...
+			Issuer:    "CerealNotes",
+		},
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	ss, err := token.SignedString(tokenSigningKey)
+	return ss, err
+}
+
+func getUserIdFromStoredToken(request *http.Request) (models.UserId, error) {
```
atmiguel: `Stored` seems like a poor choice of words here. We should reserve this term for database things

atmiguel (changes requested):

databaseutil/databaseutil.go
zingales: I thought database util was abstracting away the SQL from the function. Since UserId isn't a type in SQL, i thought it would make a better interface if it returned int64 the function that calls this one immediately casts int64 into UserId

handlers/handlers.go
zingales: If we want it want it camelCase i'll have to include it though right?

handlers/handlers.go
zingales: how about `JwtTokenClaim`

handlers/handlers.go
zingales: changing it to `StatusTemporaryRedirect` lmk if you prefer a different one.

handlers/handlers.go
zingales: Add takes the time in nanoseconds or something uselessly small. Time.Minute is the multiplier to convert from whatever time takes to minutes. I'm pretty sure this is a common go idiosyncrasy.

zingales: Just so we are clear in the future, Dates without time are `xDate` and dates with time is `xTime` or `xDateTime`

handlers/handlers.go
zingales: HttpOnly means that javascript can't access the cookie, basically a way to prevent Cross Site Scripting from stealing the cookie.

handlers/handlers.go
zingales: Well we need to delete the cookie. And since we prevent javascript from accessing the cookie using `HttpOnly:true`. I don't know how to delete it from the client side, so I think we'll need the domain to delete the cookie.

handlers/handlers.go
zingales: Good catch!

handlers/handlers.go
zingales: I think this is actually a golang cast this anonymous function into a HandlerFunc type.

handlers/handlers.go
zingales: I personally don't really like adding unnecessary periods at the end of comments. If you feel strongly about it i can add it though.

handlers/handlers.go
zingales: I forgot the convention, i actually looked through the code to find an example and thought i saw this. I'll change it to match our current convention.

handlers/handlers.go
zingales: fair i think i actually had a bug originally cause i put !=

handlers/handlers.go
zingales: I was considering moving this whole section into a different file in the same package. But wanted to talk to you about it first.

handlers/handlers.go
zingales: golang doesn't through if you have unused function parameters. I think its so that you don't have weird situation when you implement interfaces and don't use all the parameters.

handlers/handlers.go
zingales: I agree. I was waiting for you to implement the test framework so i could just throw it into whatever you made. Instead of just arbitrarily adding names or files now. I'll add a subtask to implement test framework to move this function.

main.go
zingales: The handlers are the only objects that actually use the signing key. But i'm open to suggestions

main.go
zingales: I went back and forth when i was coding and must have missed some. Changing it to `determineTokenSigningKey`.

paths/paths.go
zingales: I agree, its where i put them originally but golang doesn't allow circular imports so i had to move them to a third package. And they are currently being used in routes and handlers, and routes calls handlers

services/userservice/userservice.go
zingales: Personally i'd be a fan or removing `databaseutil.go` and incorporating all sql functions into user service. But i will change them to have the same name, though it can be kinda confusing having two functions whose api look virtually identical but do slightly different stuffs.

static/js/login_or_signup.js
zingales: nice find!

templates/home.tmpl
zingales: if that means less copy pasta-ing i'm all game. Should we make a task for it?

handlers/handlers.go
zingales: one is for the token the other is for the cookie. I gave them the same expiration to prevent sheer anarchy.

handlers/handlers.go
zingales: unfortunately you are correct.

handlers/handlers.go
zingales: makes sense.

databaseutil/databaseutil.go
atmiguel: fair enough

handlers/handlers.go
atmiguel: i think you're mixing up a couple things. I'm proposing the line look like this:
```
models.UserId `json:"userId"`
```

handlers/handlers.go
atmiguel: Just don't think about what the t in jwt stands for lol

handlers/handlers.go
atmiguel: that seems like the best fit

handlers/handlers.go
atmiguel: christ, okay

atmiguel: agreed that dates w/o time are `xDate`. My instinct for datetimes is to simply say `xTime`. But that's mainly because I've never needed to simply specify a time w/o a date.

handlers/handlers.go
atmiguel: got it. never knew about httpOnly. Only ever fucked with cookies in js

handlers/handlers.go
atmiguel: why is that necessary though? The function you're passing it has the same signature doesn't it?

handlers/handlers.go
atmiguel: Agreed, no periods if the comment is a single sentence. This comment is two sentences though

handlers/handlers.go
atmiguel: unfortunately there are situations where this inadvertently occurs (eg when a function call is inside of an if statement, before the condition). If it's confusing, we can switch to this convention always, I just think it's way uglier

handlers/handlers.go
atmiguel: A different file definitely feels right

main.go
atmiguel: Yeah, idk. I think when I made the handlers package I just expected for it to only contain request handlers. Maybe we could make a `jwtutil`?

templates/home.tmpl
atmiguel: created https://trello.com/c/EqKke9zl

handlers/handlers.go
atmiguel: Feels a bit like overkill, but it would be nice if there was a single variable we could change that would change the expiration time.

handlers/handlers.go
```diff
@@ -111,9 +111,9 @@ func HandleSessionRequest(
 			panic(err)
 		}
 
-		// Create token as a cookie and set it
+		// Set cerealNotesCookieName to have value as a our JWT Token
```
atmiguel: Sorry, I'm still not understanding this comment

handlers/handlers.go
```diff
@@ -123,57 +123,59 @@ func HandleSessionRequest(
...
-			}
+			Name:     cerealNotesCookieName,
+			Value:    "",
+			HttpOnly: true,
+			MaxAge:   -1,
+		}
 
 		http.SetCookie(responseWriter, &cookie)
 		responseWriter.WriteHeader(http.StatusOK)
 		responseWriter.Write([]byte(fmt.Sprint("user succefully logged out")))
```
atmiguel: btw, i plan on removing all the unnecessary response bodies in a later PR

databaseutil/databaseutil.go
atmiguel: rename to `GetIdForUserWithEmailAddress`

handlers/handlers.go
```diff
@@ -3,21 +3,43 @@ package handlers
...
 	"html/template"
 	"io/ioutil"
 	"log"
 	"net/http"
 	"strings"
+	"time"
 )
 
+type JwtTokenClaim struct {
+	userId models.UserId `json:"UserId"`
```
atmiguel: just a reminder to fix this line

handlers/handlers.go
```diff
@@ -3,21 +3,43 @@ package handlers
...
 // HANDLERS
 func HandleLoginOrSignupRequest(
 	responseWriter http.ResponseWriter,
 	request *http.Request,
 ) {
 	switch request.Method {
 	case http.MethodGet:
+		// Check to see if user is already logged in, if so redirect
+		if _, err := getUserIdFromJwtToken(request); err == nil {
+			http.Redirect(responseWriter, request, paths.HomePath, http.StatusTemporaryRedirect)
```
atmiguel: this line is too long

handlers/handlers.go
```diff
@@ -89,11 +111,91 @@ func HandleSessionRequest(
...
+) func(http.ResponseWriter, *http.Request) {
+	return http.HandlerFunc(
+		func(responseWriter http.ResponseWriter, request *http.Request) {
+			if userId, err := getUserIdFromJwtToken(request); err != nil {
+				// If not loggedin redirect to login page
+				http.Redirect(
+					responseWriter,
+					request,
+					paths.LoginOrSignupPath,
+					http.StatusSeeOther)
```
atmiguel: wrong http code

handlers/handlers.go
```diff
@@ -121,3 +223,73 @@ func getRequestBody(request *http.Request) []byte {
...
+		strings.TrimSpace(tokenAsString),
+		&JwtTokenClaim{},
+		func(*jwt.Token) (interface{}, error) {
+			return tokenSigningKey, nil
+		})
+}
+
+func createTokenAsString(
+	userId models.UserId,
+	expirationTimeInMinutes int64,
```
atmiguel: this name confuses me. I think you mean `minutesTilExpiration` or something?

handlers/handlers.go
```diff
@@ -121,3 +223,73 @@ func getRequestBody(request *http.Request) []byte {
...
+		return -1, err
+	}
+
+	if claims, ok := token.Claims.(*JwtTokenClaim); ok && token.Valid {
+		return claims.userId, nil
+	}
+	return -1, errors.Errorf("Token was invalid or unreadable")
+}
+
+func tokenTest1() {
```
atmiguel: reminder to move to another file

main.go
```diff
@@ -33,6 +37,19 @@ func determineDatabaseUrl() (string, error) {
 	return databaseUrl, nil
 }
 
+func determineTokenSigningKey() ([]byte, error) {
+	signingKeyVariableName := "TOKEN_SIGNING_KEY"
```
atmiguel: missing `token` in variable name

main.go
```diff
@@ -48,6 +65,16 @@ func main() {
 		}
 	}
 
+	// set up signing key
```
atmiguel: missing "token"

services/userservice/userservice.go
```diff
@@ -48,3 +49,12 @@ func AuthenticateUserCredentials(emailAddress string, password string) error {
 
 	return nil
 }
+
+func GetIdForUserWithEmailAddress(emailAddress string) (models.UserId, error) {
+	number, err := databaseutil.GetUserIdFromUserWithEmailAddress(emailAddress)
```
atmiguel: i would go with `userIdValue` as opposed to `number`

services/userservice/userservice.go
```diff
@@ -48,3 +49,12 @@ func AuthenticateUserCredentials(emailAddress string, password string) error {
...
+	if err != nil {
+		return 0, err
```
atmiguel: return -1

static/js/home.js
```diff
@@ -0,0 +1,12 @@
+$(function() {
+    $("#logout-button").click(() => {
+        $.ajax({
+            url: '/session',
+            type: 'DELETE',
+            success: function(result) {
```
atmiguel: remove unused parameter

templates/login_or_signup.tmpl
```diff
@@ -29,13 +29,13 @@
             <div class="mui-row">
                 <div class="mui-col-sm-6 mui-col-md-4 mui-col-sm-offset-3 mui-col-md-offset-4">
                     <ul class="mui-tabs__bar mui-tabs__bar--justified">
-                        <li class="mui--is-active">
+                        <li id="signup-tab">
```
atmiguel: no need for this id anymore

templates/login_or_signup.tmpl
```diff
@@ -29,13 +29,13 @@
...
                             <a data-mui-toggle="tab" data-mui-controls="signup-form">
                                 Signup
                             </a>
                         </li>
 
-                        <li>
+                        <li id="login-tab" class="mui--is-active">
```
atmiguel: no need for this id

atmiguel (changes requested):

handlers/handlers.go
zingales: I tried compiling it without the cast and it works. I got this code from a website and I think this is in case a function caller requires a http.HanderFunc type. Since go type checks even if they are the same type, it won't implicitly "upcast" a type. So for example if the function demanded an input as UserIdType and you gave it an int64 it would throw an error unless you cast it, however, if the function required an int64 and you gave it a UserId i don't think it would throw an error. I mild preference for leaving it, but can be easily convinced to remove it.

handlers/handlers.go
zingales: I'll let you decide on this one.

main.go
zingales: I'm going to leave it in this package for now until we encounter this problem again. I already feel like we have too many packages that only do one thing.

handlers/handlers.go
zingales: previous comment says I'll do it after you create testing framework

services/userservice/userservice.go
zingales: used `userIdAsInt`

templates/login_or_signup.tmpl
zingales: are you against having ids we don't use?

handlers/handlers.go
zingales: ask and thee shall receive

templates/login_or_signup.tmpl
atmiguel: yes

atmiguel: to me it's no different from leaving unused variables all over. If i'm reading through this html and I see that id, I assume it's being used in the css or js.

atmiguel (approved):

PR #15, opened 2018-03-30
Email wrapper

zingales:
Guarantees that EmailAddress.String is lower case.


models/user.go
```diff
@@ -1,3 +1,17 @@
...
+package models
 
-type UserId int64
\ No newline at end of file
+import "strings"
+
+type UserId int64
+
+type EmailAddress struct {
+	emailAsString string
```
atmiguel: would prefer `emailAddressAsString`

models/user.go
```diff
@@ -1,3 +1,17 @@
...
+}
+
+func NewEmailAddress(emailAddr string) *EmailAddress {
```
atmiguel: why `emailAddr`? would prefer a long variable name

models/user.go
```diff
@@ -1,3 +1,17 @@
-package models 
+package models
 
-type UserId int64
\ No newline at end of file
+import "strings"
+
+type UserId int64
+
+type EmailAddress struct {
```
atmiguel, line 7: Seems weird for this to be in `user.go`. Any reason not to create `email.go` or something?

handlers/handlers.go
```diff
@@ -100,6 +100,30 @@ func HandleUserRequest(
 	}
 }
 
+func HandleRootRequest(
+	responseWriter http.ResponseWriter,
+	request *http.Request,
+) {
+	switch request.Method {
+	case http.MethodGet:
+		if request.URL.Path == "/" {
```
atmiguel: why this? didn't we say we'd just have all "not founds" redirect to home for now? i could be remembering wrong though

routers/routers.go
```diff
@@ -19,6 +19,9 @@ func SetRoutes() {
 			http.StripPrefix(staticDirectoryPaddedWithSlashes, fileServer))
 	}
 
+	// Root path will redirect to home
+	http.HandleFunc("/", handlers.HandleRootRequest)
```
atmiguel: not sure how i feel about it, but we could maybe just name the function something like `redirectGetRequestToHome`?

atmiguel (changes requested):

models/user.go
zingales: I'll do emailAddressAsString then. Didn't want to use emailAddress lower case, especially considering its not an instance of the EmailAddress the type.

routers/routers.go
zingales: Out of curiosity why have a specific functionality name for this handle request, but not for other ones?

handlers/handlers.go
zingales: Your right we did talk about that, I'll do that, though i want to double check your okay with this "wonky" behavior. 

For example since we have `/user` handler,  `/user/banana` will result in HTTP.NotFound exception. However, `/bob/banana` will redirect home page since there is no `/bob` handler

models/user.go
zingales: I consider UserId and Email address to be part of the User model.

atmiguel: Hmm, yeah i might be generalizing too quickly again. As of now, what you're saying is correct, so yeah let's leave it here

atmiguel (approved):

PR #16, opened 2018-03-31
Update routers.go

ethomas2:


ethomas2:
Minor, but you guys didn't run `gofmt` on the most recent commit of your code. I don't really care about `gofmt`, in fact I disagree with a lot of the time, but I know some folks that think it's nice for code to be consistent across the go community.
` find . -iname \*go -exec gofmt -w {} \;`

or for extra points add a `-s`
` find . -iname \*go -exec gofmt -s -w {} \;`

routers/routers.go
```diff
@@ -6,7 +6,7 @@ import (
 	"net/http"
 )
 
-func SetRoutes() {
+func SetRoutes() {
```
ethomas2: I'm not a big fan of `SetRoutes()` setting global state. In main.go you call `http.ListenAndServe(port, nil)`, and it's not easy for me to figure out where the handler is coming from. I'd prefer something like
```go
// In routers.go
func GetRoutes() http.Handler {
  mux := http.NewServeMux()
  mux.HandleFunc(paths.LoginOrSignupPath, handlers.HandleLoginOrSignupRequest)
  mux.HandleFunc(paths.LoginOrSignupPath, handlers.HandleLoginOrSignupRequest)
  mux.HandleFunc("/user", handlers.HandleUserRequest)
  mux.HandleFunc("/session", handlers.HandleSessionRequest)
  return mux
}

// In main.go
func main() {
  handler := routers.GetRoutes()
  http.ListenAndServe(port, handler)
}
```

GetRoutes might be a terrible name, but you get the idea

routers/routers.go
```diff
@@ -1,38 +1,38 @@
...
-}
+package routers 
+ 
+import ( 
+	"github.com/atmiguel/cerealnotes/handlers" 
+	"github.com/atmiguel/cerealnotes/paths" 
+	"net/http" 
+) 
+ 
+func SetRoutes() {
```
ethomas2, line 9: I'm not a big fan of `SetRoutes()` setting global state. In main.go you call `http.ListenAndServe(port, nil)`, and it's not easy for me to figure out where the handler is coming from. I'd prefer something like
```go
// In routers.go
func GetRoutes() http.Handler {
  mux := http.NewServeMux()
  mux.HandleFunc(paths.LoginOrSignupPath, handlers.HandleLoginOrSignupRequest)
  mux.HandleFunc(paths.LoginOrSignupPath, handlers.HandleLoginOrSignupRequest)
  mux.HandleFunc("/user", handlers.HandleUserRequest)
  mux.HandleFunc("/session", handlers.HandleSessionRequest)
  mux.HandleFunc(paths.HomePath, handlersAuthenticateOrRedirectToLogin(handlers.HandleHomeRequest))
  ...
  return mux
}

// In main.go
func main() {
  handler := routers.GetRoutes()
  http.ListenAndServe(port, handler)
}
```

GetRoutes might be a terrible name, but you get the idea

handlers/handlers.go
```diff
@@ -1,238 +1,238 @@
...
+				Value:    token, 
+				Expires:  expirationTime, 
+				HttpOnly: true, 
+			} 
+ 
+			http.SetCookie(responseWriter, &cookie) 
+		} 
+ 
+		responseWriter.WriteHeader(http.StatusCreated) 
+		responseWriter.Write([]byte(fmt.Sprint("passward email combo was correct")))
```
ethomas2, line 153: `responseWriter` is a `io.Writer`, which means you can do
`fmt.Fprint(responseWriter, "password email combo was correct")`

there's also a `fmt.Fprintln` and `fmt.Fprintf` in case you ever need that

handlers/handlers.go
```diff
@@ -1,238 +1,238 @@
...
+) { 
+	allowedMethodsString := strings.Join(allowedMethods, ", ") 
+	responseWriter.Header().Set("Allow", allowedMethodsString) 
+ 
+	statusCode := http.StatusMethodNotAllowed 
+	http.Error(responseWriter, http.StatusText(statusCode), statusCode) 
+} 
+ 
+func getRequestBody(request *http.Request) []byte { 
+	body, err := ioutil.ReadAll(request.Body)
```
ethomas2, line 228: This line is reading all the bytes from the `io.Reader` into memory. When possible, avoid reading all the bytes from an `io.Reader` into memory, instead give the `io.Reader` directly to whatever wants to consume it. That way the consumer can decide if it needs to read the whole byte stream, and even if it does need to read the whole byte stream it can decide if it needs all the bytes in memory at the same time or if it can stream it.

It looks like the bytes from `getRequestBody` are only being consumed by `json.Unmarshal`. I'd get rid of `getRequestBody()` and do the  so instead you could do

```go
func HandleUserRequest(...) {
   defer func() {
       io.CopyN(ioutil.Discard, request.Body, 512) 
       request.Body.Close()
   }()
   
   ...

   signupForm := &SignupForm{}  // alternatively signupForm := new(SignupForm)
   if err := json.NewDecoder(request.Body).Decode(signupForm); err != nil {
       ...
   }
}
```

ethomas2: Edit: Ignore everything below. Apparently as of go1.7 you don't need to drain the `request.Body`. 
See https://github.com/google/go-github/pull/847
and https://github.com/golang/go/commit/18072ad

~~As a side note, i'm not 100% sure why you want to drain the reader before closing it. Someone told me to do it once. I've also seen this in go-github, which is a github client written in go that we use at work. I really like their code, and I copy a lot of their patterns.~~ 

~~I chose the number 512 because that's what go-github does. I'm not sure why they do that and tbh 512 seems pretty low. I think it probably makes more sense to do `io.Copy(ioutil.Discard, request.Body)`, idk why go-github doesn't do that.~~

~~I _think_ the reason you want to drain the request.Body is because the drain will wait for the guy on the other end of the tcp connection to finish sending all of his data before terminating the tcp connection. If you Close before draining you risk closing the connection before the other guy is done.~~

~~Actually, now that I write this out, that's totally why you want to drain before closing.~~

~~It's worth noting that in this example, the drain is technically unnecessary since you're always calling `json.NewDecoder.Decode` on `request.Body` which will drain it for you ... but that's an implementation detail of the json package that you shouldn't depend on.~~

handlers/handlers.go
```diff
@@ -1,238 +1,238 @@
...
+		Password     string `json:"password"` 
+	} 
+ 
+	switch request.Method { 
+	case http.MethodPost: 
+		var signupForm SignupForm 
+		body := getRequestBody(request) 
+ 
+		if err := json.Unmarshal(body, &signupForm); err != nil { 
+			panic(err)
```
ethomas2, line 78: These panics are pretty bad
1. panic doesn't just kill the current goroutine, it crashes the whole program. So one user's malformed request could crash your whole server instead of just 400ing to that user. Be very wary of panics. panic should only be used in cases where the only reasonable thing to do is crash the whole service. I've only ever used panic very early in the startup of a service (e.g. failure to load the config file), or in places where something is logically impossible (e.g. in the default of a switch statement that should never occur).

2. You _could_ [defer a recover](https://blog.golang.org/defer-panic-and-recover) at the top of this function. That would work, but it would be non-idomatic. The idiomatic thing to do is to immediately handle the error and return, or bubble the error up to the caller (in this case there is no caller). Obvs if the error handling gets complicated or duplicated a bunch, then factor the error handling out into its own function

databaseutil/databaseutil.go
```diff
@@ -1,105 +1,105 @@
...
+	displayName string, 
+	emailAddress string, 
+	password []byte, 
+	creationTime time.Time, 
+) error { 
+	var row *sql.Row 
+	{ 
+		sqlQuery := ` 
+			INSERT INTO users (display_name, email_address, password, creation_time) 
+			VALUES ($1, $2, $3, $4)`
```
ethomas2, line 42: sql injection vulnerability here. You're probably aware of this already and just don't care (I wouldn't care if I were you, it's not like this is a production service). But in the interest of leaving a good code review, I feel like I should note this.

ethomas2: More sql injections later on in the file

databaseutil/databaseutil.go
```diff
@@ -1,105 +1,105 @@
...
+		sqlQuery := ` 
+			SELECT id FROM users 
+			WHERE email_address = $1` 
+ 
+		row = db.QueryRow(sqlQuery, emailAddress) 
+	} 
+ 
+	var userId int64 
+	if err := row.Scan(&userId); err != nil { 
+		return -1, err
```
ethomas2, line 88: Usually when a function returns two values `(foo, error)`, if `error` is non-nil, `foo` is the [zero value](https://tour.golang.org/basics/12). The zero value for numeric types is 0, not -1

handlers/handlers.go
```diff
@@ -1,238 +1,238 @@
...
+	"github.com/dgrijalva/jwt-go" 
+	"html/template" 
+	"io/ioutil" 
+	"log" 
+	"net/http" 
+	"strings" 
+	"time" 
+) 
+ 
+const oneWeekInMinutes = 60 * 24 * 7
```
ethomas2, line 18: Probably makes more sense for this to be a [time.Duration](https://golang.org/pkg/time/#Duration) instead of an int.

`const oneWeek = time.Hour * 24 * 7`

You can get the number of minutes that this duration is by doing `duration.Minutes()`.
Using `time.Duration` gives you the standard advantages that a time library gives you. Lets you do time arithmetic, lets you see if one time comes before or after another, better formatting when printing times ... stuff like that

ethomas2: When you end up using this in`tokenutil`, you do something like
```go
jwt.StandardClaims{
   ExpiresAt: time.Now().Add(oneWeek).Unix()
}
```

handlers/handlers.go
```diff
@@ -1,238 +1,238 @@
...
+ 
+var tokenSigningKey []byte 
+ 
+func SetTokenSigningKey(key []byte) { 
+	tokenSigningKey = key 
+} 
+ 
+// HANDLERS 
+func HandleLoginOrSignupRequest( 
+	responseWriter http.ResponseWriter,
```
ethomas2, line 35: In general I completely agree with you guys that single letter variable names are a cancer, but I do use the signature `(w http.ResponseWriter, r  *http.Request)` a lot. This function signature is _super_ common. This is the rare exception where I think spelling out a descriptive variable name is less readable than using the single letter. Just because i'm so used to seeing `(w http.ResponseWriter, r  *http.Request)`, when I see that signature I immediately pattern match and think "oh, it's one of _those_ types of functions".

handlers/handlers.go
```diff
@@ -1,238 +1,238 @@
...
+				responseWriter, 
+				request, 
+				paths.HomePath, 
+				http.StatusTemporaryRedirect) 
+			return 
+		} 
+ 
+		parsedTemplate, err := template.ParseFiles("templates/login_or_signup.tmpl") 
+		if err != nil { 
+			log.Fatal(err)
```
ethomas2, line 52: This is similar to panic in that it's gonna crash the whole program instead of returning an error status for this one request
```
mbp-IT002597:go ethomas$ go doc log.Fatal
func Fatal(v ...interface{})
    Fatal is equivalent to Print() followed by a call to os.Exit(1).

```

handlers/handlers.go
```diff
@@ -1,238 +1,238 @@
...
+			[]string{http.MethodPost, http.MethodDelete}) 
+	} 
+} 
+ 
+type AuthentictedRequestHandlerType func( 
+	http.ResponseWriter, 
+	*http.Request, 
+	models.UserId) 
+ 
+func AuthenticateOrRedirectToLogin(
```
ethomas2, line 179: omg I love this function so much. A function that takes a function and returns another function. This smells so much of haskell

ethomas2: I think I might adopt this pattern for authenticated endpoints at work

ethomas2: Minor: The return value of this function is 
`func(http.ResponseWriter, *http.Request)`
you could change it to
`http.HandlerFunc`, which is the same thing, but is more concise and possibly more readable (there's an argument to be made that it's no more readable. So ... whatevs)

databaseutil/databaseutil.go
ethomas2: Same for a bunch of methods in tokenutil

zingales: Is there a zero value function? so we could just do return int.ZeroValue, err. Cause that would be more readable.

ethomas2: don't think so, I think you just have to memorize them. But they're all what you'd expect. Just imagine that type in memory with every byte being zero.

ethomas2: ah I see, you're concerned about readability, not memorability. Still, the answer is no

handlers/handlers.go
zingales: error: handlers/handlers.go:202:3: http.HandleFunc is not a type

databaseutil/databaseutil.go
zingales: I looked into this and i think sql.Query and sql.QueryRow are injection safe.

handlers/handlers.go
atmiguel: @ethomas2 could you explain your code snippet above? I don't understand what's happening in the deferred function

databaseutil/databaseutil.go
atmiguel: we def have a task to clean all user input

handlers/handlers.go
atmiguel: Totally understand that, but i had a really hard time reading all the code i found written like that. Like damn one of the variables is `r` and i was always like wait is this the request or the response. I understand what you mean about getting used to things, but I often will `index` over `i` in for loops, as well as `event` over `e` in js event handlers. Just makes more sense to me

handlers/handlers.go
ethomas2: It takes up to 512 bytes from the request body and sends it to the equivilent of /dev/null. 

I used 512 instead of just discarding all the bytes because that's what go-github does. It probably makes more sense to discard all the bytes with `io.Copy`

All of that is irrelevant though. Apparently you don't have to drain the request body after go1.7

handlers/handlers.go
ethomas2: HandlerFunc, not HandleFunc

databaseutil/databaseutil.go
zingales: @atmiguel how do you feel about `-1` vs `0`. Cause i think we specifically changed it from 0 -> -1

PR #18, opened 2018-04-01
Addressed Evans comments

zingales:
Except return -1, nil.

### DIFF
--- a/databaseutil/databaseutil.go
+++ b/databaseutil/databaseutil.go
@@ -36,24 +36,21 @@ func InsertIntoUsersTable(
 	password []byte,
 	creationTime time.Time,
 ) error {
-	var row *sql.Row
-	{
-		sqlQuery := `
-			INSERT INTO users (display_name, email_address, password, creation_time)
-			VALUES ($1, $2, $3, $4)`
+	sqlQuery := `
+		INSERT INTO users (display_name, email_address, password, creation_time)
+		VALUES ($1, $2, $3, $4)`
 
-		row = db.QueryRow(sqlQuery, displayName, emailAddress, password, creationTime)
+	rows, err := db.Query(sqlQuery, displayName, emailAddress, password, creationTime)
+	if err != nil {
+		return convertPostgresError(err)
 	}
+	defer rows.Close()
 
-	if err := row.Scan(); err != nil {
-		if err == sql.ErrNoRows { // success
-			return nil
-		}
-
+	if err = rows.Err(); err != nil {
 		return convertPostgresError(err)
 	}
 
-	panic("unexpected")
+	return nil
 }
 
 func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
--- a/databaseutil/databaseutil.go
+++ b/databaseutil/databaseutil.go
@@ -46,7 +46,7 @@ func InsertIntoUsersTable(
 	}
 	defer rows.Close()
 
-	if err = rows.Err(); err != nil {
+	if err := rows.Err(); err != nil {
 		return convertPostgresError(err)
 	}
 
--- a/handlers/handlers.go
+++ b/handlers/handlers.go
@@ -4,8 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
-	"io/ioutil"
-	"log"
 	"net/http"
 	"strings"
 	"time"
@@ -16,8 +14,8 @@ import (
 	"github.com/dgrijalva/jwt-go"
 )
 
-const oneWeekInMinutes = 60 * 24 * 7
-const credentialTimeoutInMinutes = oneWeekInMinutes
+const oneWeek = time.Hour * 24 * 7
+const credentialTimeoutDuration = oneWeek
 const cerealNotesCookieName = "CerealNotesToken"
 
 type JwtTokenClaim struct {
@@ -50,7 +48,8 @@ func HandleLoginOrSignupRequest(
 
 		parsedTemplate, err := template.ParseFiles("templates/login_or_signup.tmpl")
 		if err != nil {
-			log.Fatal(err)
+			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+			return
 		}
 
 		parsedTemplate.Execute(responseWriter, nil)
@@ -72,11 +71,11 @@ func HandleUserRequest(
 
 	switch request.Method {
 	case http.MethodPost:
-		var signupForm SignupForm
-		body := getRequestBody(request)
+		signupForm := new(SignupForm)
 
-		if err := json.Unmarshal(body, &signupForm); err != nil {
-			panic(err)
+		if err := json.NewDecoder(request.Body).Decode(signupForm); err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+			return
 		}
 
 		var statusCode int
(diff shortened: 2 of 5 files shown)


routers/routers.go
```diff
@@ -7,35 +7,39 @@ import (
 	"github.com/atmiguel/cerealnotes/paths"
 )
 
-func SetRoutes() {
+func DefineRoutes() http.Handler {
```
zingales, line 10: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178422483

handlers/handlers.go
```diff
@@ -171,7 +174,7 @@ func HandleSessionRequest(
 		}
 
 		responseWriter.WriteHeader(http.StatusCreated)
-		responseWriter.Write([]byte(fmt.Sprint("passward email combo was correct")))
+		fmt.Fprint(responseWriter, "passward email combo was correct")
```
zingales, line 177: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178422704

handlers/handlers.go
```diff
@@ -244,16 +248,3 @@ func respondWithMethodNotAllowed(
 	statusCode := http.StatusMethodNotAllowed
 	http.Error(responseWriter, http.StatusText(statusCode), statusCode)
 }
-
```
zingales, line 247: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178423405

handlers/handlers.go
```diff
@@ -131,34 +131,37 @@ func HandleSessionRequest(
...
 	switch request.Method {
 	case http.MethodPost:
-		var loginForm LoginForm
-		body := getRequestBody(request)
+		loginForm := new(LoginForm)
 
-		if err := json.Unmarshal(body, &loginForm); err != nil {
-			panic(err)
+		if err := json.NewDecoder(request.Body).Decode(loginForm); err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
```
zingales, line 137: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178423894

databaseutil/databaseutil.go
```diff
@@ -36,24 +36,21 @@ func InsertIntoUsersTable(
 	password []byte,
 	creationTime time.Time,
 ) error {
-	var row *sql.Row
-	{
-		sqlQuery := `
-			INSERT INTO users (display_name, email_address, password, creation_time)
-			VALUES ($1, $2, $3, $4)`
+	sqlQuery := `
```
zingales, line 39: Started looking at this code due to: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178424005 and realized we shouldn't be using the query row here since it requires a row so i was able to fix the code by using query instead.

handlers/handlers.go
```diff
@@ -199,7 +202,7 @@ type AuthentictedRequestHandlerType func(
 
 func AuthenticateOrRedirectToLogin(
 	authenticatedHandlerFunc AuthentictedRequestHandlerType,
-) func(http.ResponseWriter, *http.Request) {
+) http.HandlerFunc {
```
zingales, line 205: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178424911

handlers/handlers.go
```diff
@@ -16,8 +14,8 @@ import (
 	"github.com/dgrijalva/jwt-go"
 )
 
-const oneWeekInMinutes = 60 * 24 * 7
-const credentialTimeoutInMinutes = oneWeekInMinutes
+const oneWeek = time.Hour * 24 * 7
```
zingales, line 17: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178424409

handlers/tokenutil.go
```diff
@@ -23,12 +23,12 @@ func parseTokenFromString(tokenAsString string) (*jwt.Token, error) {
...
 func createTokenAsString(
 	userId models.UserId,
-	minutesTilExpiration int64,
+	durationTilExpiration time.Duration,
 ) (string, error) {
 	claims := JwtTokenClaim{
 		userId,
 		jwt.StandardClaims{
-			ExpiresAt: time.Now().Unix() + (minutesTilExpiration * 60),
+			ExpiresAt: time.Now().Add(durationTilExpiration).Unix(),
```
zingales, line 31: https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178424467

ethomas2:
👍

ethomas2 (approved):

databaseutil/databaseutil.go
atmiguel: what do you mean by "it requires a row"?

databaseutil/databaseutil.go
```diff
@@ -36,24 +36,21 @@ func InsertIntoUsersTable(
...
+		INSERT INTO users (display_name, email_address, password, creation_time)
+		VALUES ($1, $2, $3, $4)`
 
-		row = db.QueryRow(sqlQuery, displayName, emailAddress, password, creationTime)
+	rows, err := db.Query(sqlQuery, displayName, emailAddress, password, creationTime)
+	if err != nil {
+		return convertPostgresError(err)
 	}
+	defer rows.Close()
```
atmiguel, line 47: why is this necessary? and why wasn't it necessary before?

atmiguel, line 47: Shouldn't this be done before the error check?

databaseutil/databaseutil.go
```diff
@@ -36,24 +36,21 @@ func InsertIntoUsersTable(
...
 
-	if err := row.Scan(); err != nil {
-		if err == sql.ErrNoRows { // success
-			return nil
-		}
-
+	if err = rows.Err(); err != nil {
```
atmiguel: random nitpicky thing, but we should do `:=` here. Honestly wish `err` wasn't in scope here, so this couldn't happen, but that doesn't seem worth the hassle.

handlers/handlers.go
```diff
@@ -131,34 +131,37 @@ func HandleSessionRequest(
...
-				panic(err)
+				http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+				return
 			}
 
-			token, err := createTokenAsString(userId, credentialTimeoutInMinutes)
+			token, err := createTokenAsString(userId, credentialTimeoutDuration)
 			if err != nil {
-				panic(err)
+				http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
```
atmiguel, line 160: this is the sort of thing we'd ideally handle in some sort of middleware (seeing how often you've had to write it), but this is perfectly good for now

atmiguel (changes requested):

databaseutil/databaseutil.go
zingales: i meant it doesn't require a row. QueryRow expects a single row response. An insert does not return rows, and that's why before we had to check for if ErrNoRows, but now i don't need to do that check.

databaseutil/databaseutil.go
zingales: I actually did that  originally, but it resulted in an invalid pointer check, because if the first error returns then the rows object is nil. So its only after you know that err == nil that you now rows is defined.

zingales: Honestly it might have been necessary before. I just copied what i saw online when dealing with rows.

handlers/handlers.go
zingales: Ya i agree, i just didn't know how to write the middleware and and i was like one problem at a time.

databaseutil/databaseutil.go
zingales: I can double check but i think its because err is already defined from the previous statement so i can't := it here.

atmiguel: I just double-checked. That seems incorrect. Go doesn't have an issue with redefining non-const variables.

atmiguel (approved):

PR #20, opened 2018-04-06
Better redirection

zingales:
* created a favicon
* redirected a /favicon to /static/favicon
* throw an error when invalid password is given and show an alert
* pressing enter on login now submits the form
* created a path constant for all non-redirect paths


services/userservice/userservice.go
```diff
@@ -17,6 +17,10 @@ import (
 // cannot be used becaues it already being used by another user
 var EmailAddressAlreadyInUseError = errors.New("Email address already in use")
 
+// InvalidPasswordError is returned when the password provided does not match
+// the password we have in our system
+var InvalidPasswordError = errors.New("The password did not match")
```
atmiguel: `Invalid` seems like the wrong word. The password was likely valid, it just didn't match.

atmiguel: Also, we need to return some named error when we don't recognize the email address. We could simply have a `CredentialsNotAuthorizedError` or something like that.

handlers/handlers.go
```diff
@@ -159,7 +138,14 @@ func HandleSessionRequest(
 			models.NewEmailAddress(loginForm.EmailAddress),
 			loginForm.Password,
 		); err != nil {
-			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+			statusCode := http.StatusInternalServerError
+			if err == userservice.InvalidPasswordError {
+				statusCode = http.StatusUnauthorized
+			}
+			http.Error(
```
atmiguel: can all be on one line

handlers/handlers.go
```diff
@@ -159,7 +138,14 @@ func HandleSessionRequest(
 			models.NewEmailAddress(loginForm.EmailAddress),
 			loginForm.Password,
 		); err != nil {
-			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+			statusCode := http.StatusInternalServerError
+			if err == userservice.InvalidPasswordError {
```
atmiguel: would prefer an if else for this logic, but it's nbd

handlers/handlers.go
```diff
@@ -232,14 +218,36 @@ func AuthenticateOrRedirectToLogin(
...
 				http.StatusTemporaryRedirect)
 		} else {
 			authenticatedHandlerFunc(responseWriter, request, userId)
 		}
 	}
 }
 
+// GetRedirectHandler is a function which given a path returns a handler that
+// on get requests redirects to the given path.
+func GetRedirectHandler(
```
atmiguel: should probably say something about where it's redirecting to. Something like `HandleRedirectToPath`

handlers/handlers.go
```diff
@@ -232,14 +218,36 @@ func AuthenticateOrRedirectToLogin(
...
+				paths.LoginOrSignup,
 				http.StatusTemporaryRedirect)
 		} else {
 			authenticatedHandlerFunc(responseWriter, request, userId)
 		}
 	}
 }
 
+// GetRedirectHandler is a function which given a path returns a handler that
+// on get requests redirects to the given path.
```
atmiguel: We still need to talk about godoc comments but if we decide that these stay, we should at least make them as concise as possible.

Never need to say `is a function`.
No need to say `given a path` twice.
`get` should be all caps.

handlers/handlers.go
```diff
@@ -232,14 +218,36 @@ func AuthenticateOrRedirectToLogin(
...
+func GetRedirectHandler(
+	newPath string,
```
atmiguel: `newPath` is weird in the context of this function, because there's no original path to differentiate against

static/js/login_or_signup.js
```diff
@@ -98,7 +102,8 @@ $(function() {
     attachFieldValidators(loginFormMetadata);
 
     var attachSubmitClickHandler = function(formMetadata, postFunction) {
-        formMetadata.$form.find('button').click(() => {
+        var button = formMetadata.$form.find('button');
```
atmiguel: should be `$button`

static/js/login_or_signup.js
```diff
@@ -118,6 +123,12 @@ $(function() {
                 touchAllFields(formMetadata.$form, formMetadata.fields);
             }
         });
+
+        formMetadata.$form.find("input:password").keyup((event) => {
```
atmiguel: single quotes

atmiguel: you did `keyup` instead of `keypress`. Can you explain to me why? I forget the difference.

static/js/login_or_signup.js
```diff
@@ -144,6 +155,12 @@ $(function() {
             } else {
                 alert('Error in logging in');
             }
-        }, 'text');
+        }, 'text').fail((request) => {
+            if (request.status === 401) {
+                alert("Password was not correct");
```
atmiguel: single quotes

atmiguel (changes requested):

handlers/handlers.go
zingales: I wanted to make sure that it naming didn't seem to imply that it was a handler itself, that's why i didn't put Handle as the begging of the name. How about `RedirectToPathHandler`?

handlers/handlers.go
zingales: We decided they don't

handlers/handlers.go
zingales: `finalPath`?

static/js/login_or_signup.js
zingales: actually reading up on the difference i think it should be `keydown`. The difference between the 3 (up, down and press) is weather or not the character for that key even is inserted into the "text" field: down is before, press is during and up is after.  

but to answer your original question, i think it's how i saw it done online.

services/userservice/userservice.go
zingales: No longer needed.

handlers/handlers.go
atmiguel: modifying it with `final` is still confusing when there's no non-final path in context. Are you opposed to simply naming it `path`?

atmiguel (approved):

PR #21, opened 2018-04-07
Dry Base tmpl

zingales:
Created a base template to make our webpages DRYer

### DIFF
--- a/templates/base.tmpl
+++ b/templates/base.tmpl
@@ -1,0 +1,28 @@
+{{ define "base" }}
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8" />
+        <meta
+            name="description"
+            content="A better way to book club" />
+        <meta
+            name="viewport"
+            content="width=device-width, initial-scale=1" />
+
+        <title>{{block "title" .}} {{end}}</title>
+
+        <!-- CSS -->
+        <link href="//cdn.muicss.com/mui-0.9.36/css/mui.min.css" rel="stylesheet" type="text/css" />
+        {{block "css" .}} {{end}}
+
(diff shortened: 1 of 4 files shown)


templates/base.tmpl
```diff
@@ -0,0 +1,28 @@
...
+    <head>
+        <meta charset="utf-8" />
+        <meta
+            name="description"
+            content="A better way to book club" />
+        <meta
+            name="viewport"
+            content="width=device-width, initial-scale=1" />
+
+        <title>CerealNotes | {{block "title" .}} {{end}}</title>
```
atmiguel: We should be consistent about our `{{ }}` spacing.
I prefer spaces:
```{{ block "title" . }}```

File: templates/home.tmpl [hunk 2]
```diff
@@ -1,33 +1,16 @@
...
-        <meta
-            name="description"
-            content="A better way to book club" />
-        <meta
-            name="viewport"
-            content="width=device-width, initial-scale=1" />
+{{define "js"}}<script src="/static/js/home.js"></script>{{end}}
 
-        <title>CerealNotes</title>
+{{define "content"}}
```
atmiguel: we should be consistent about tabbing after a `define`. I prefer what you did here where it's tabbed, but in base.tmpl it's not tabbed

atmiguel (changes requested):

atmiguel (changes requested):
Just reiterating this old comment:
We should be consistent about our `{{ }}` spacing.
I prefer spaces:
```{{ block "title" . }}```

Looks like you made the change in only one spot. Should be done everywhere with `{{ }}` syntax

atmiguel (approved):

PR #23, opened 2018-04-10
Cleaned up allowed methods function and error return values

atmiguel:


handlers/handlers.go
```diff
@@ -246,16 +245,20 @@ func HandleHomeRequest(
...
+		respondWithMethodNotAllowed(responseWriter, http.MethodGet)
 	}
 }
 
 // PRIVATE
 
 func respondWithMethodNotAllowed(
 	responseWriter http.ResponseWriter,
-	allowedMethods []string,
+	allowedMethods ...string,
```
zingales, line 256: I like this very slick.

handlers/handlers.go
```diff
@@ -246,16 +245,20 @@ func HandleHomeRequest(
...
 ) {
+	if allowedMethods == nil {
+		return
```
zingales: Shouldn't this be an error? What is a scenario where calling this method without inputs makes sense?

But this function also might make sense to go into our middle layer if we ever get around to it.

handlers/tokenutil.go
```diff
@@ -40,18 +40,19 @@ func createTokenAsString(
...
-		return -1, err
+		return 0, err
 	}
 
 	if claims, ok := token.Claims.(*JwtTokenClaim); ok && token.Valid {
 		return claims.UserId, nil
 	}
-	return -1, errors.New("Token was invalid or unreadable")
+
+	return 0, errors.New("Token was invalid or unreadable")
```
zingales: I know this is code i wrote, but I'm realizing we its actually more annoying to return a new error every time rather than a public var error that is defined in this class.

zingales (changes requested):

handlers/tokenutil.go
atmiguel: what do you mean by "more annoying"? you mean in terms of handling the error?

handlers/handlers.go
atmiguel: Yeah i suppose you're right. What kind of error makes sense here? Should I panic?

handlers/tokenutil.go
zingales: Ya. I mean it makes error handling harder, since in the calling function, its hard to check if error == this specific error, and you'd have to check if the text is what you expect.

handlers/handlers.go
zingales: So the dream (as far as i know) is not achievable, dream being that its a compiler error. This is one of those places where i would dream of having different behavior in dev and in prod. In prod i want it to just return a generic method, something like Method passed in was not allowed. In Dev i probably want it to panic. Since code should never be checked in with this allowed method == nil.

zingales (approved):

handlers/handlers.go
ethomas2: I'm in here. Lurking in your code reviews

You could make this a compiler error by making the signature

```go
func respondWithMethodNotAllowed(
  allowedMethod string, 
  additionalMethods  ...string)
```

then maybe the first line of this method is to concatenate the required string with the  optional strings

PR #24, opened 2018-06-03
Notes page

atmiguel:


models/note.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	MARGINALIA NoteType = iota
+	META
+	QUESTIONS
+	PREDICTIONS
+)
+
+func (noteType NoteType) String() string {
+	types := [...]string{
```
zingales: nit: this should probably be outside the function, as an internal package variable, so it doesn't get initialized every time you call the function.

handlers/handlers.go
```diff
@@ -75,6 +75,25 @@ func HandleUserRequest(
 	}
 
 	switch request.Method {
+	case http.MethodGet:
```
zingales, line 78: Why do we need to get all the usernames?

handlers/handlers.go
```diff
@@ -166,20 +186,22 @@ func HandleSessionRequest(
 		}
 
 		responseWriter.WriteHeader(http.StatusCreated)
-		fmt.Fprint(responseWriter, "passward email combo was correct")
+		fmt.Fprint(responseWriter, "password email combo was correct")
```
zingales: are there any conventions of what non web-page api's should return?

models/note.go
```diff
@@ -0,0 +1,52 @@
...
+	Type          NoteType      `json:"type"`
+	Content       string        `json:"content"`
+	PublicationId PublicationId `json:"publicationId"`
+	CreationTime  time.Time     `json:"creationTime"`
+}
+
+func (note *Note) MarshalJSON() ([]byte, error) {
+	type Alias Note
+
+	return json.Marshal(&struct {
```
zingales, line 45: I don't understand what this does. It converts a Note into a []byte? Why doe we need the Alias type?

routers/routers.go
```diff
@@ -26,18 +26,19 @@ func DefineRoutes() http.Handler {
...
 
-	// forms
-	mux.HandleFunc(paths.User, handlers.HandleUserRequest)
-	mux.HandleFunc(paths.Session, handlers.HandleSessionRequest)
+	handleAuthenticated(mux, paths.HomePage, handlers.HandleHomePageRequest)
+	handleAuthenticated(mux, paths.NotesPage, handlers.HandleNotesPageRequest)
 
-	// requires authentication
-	handleAuthenticated(mux, paths.Home, handlers.HandleHomeRequest)
+	// ajax
```
zingales: Why ajax and not api?

static/js/login_or_signup.js
```diff
@@ -149,7 +149,7 @@ $(function() {
     });
 
     attachSubmitClickHandler(loginFormMetadata, (formDataAsJsonString) => {
-        $.post('/session', formDataAsJsonString, (responseBody, _, request) => {
+        $.post('/api/session', formDataAsJsonString, (responseBody, _, request) => {
```
zingales, line 152: it sucks that there is no way to export all the urls in that we have in our paths, so that they can be accessed through javascript.

static/js/notes.js
```diff
@@ -0,0 +1,55 @@
+var USERS_BY_ID = {};
+
+const $createAuthor = function(authorId) {
```
zingales, line 3: Nit: these createX should probably be more like createXDom or createXHtml.

static/js/notes.js
```diff
@@ -0,0 +1,55 @@
...
+        .append($creationTime);
+
+    return $('<div>').addClass('note')
+        .append($header)
+        .append($content);
+};
+
+$(function() {
+    $.get('/api/user', function(usersById) {
+        USERS_BY_ID = usersById;
```
zingales, line 43: If i understand correctly, we need this because the note object stores a userId and not a user Display Name. So we need to convert userId's to User?

And normally we'd convert that on the database side, however, that could lead to mistakes if different users had the same display name?

handlers/handlers.go
```diff
@@ -75,6 +75,25 @@ func HandleUserRequest(
 	}
 
 	switch request.Method {
+	case http.MethodGet:
+		user1 := models.User{"Adrian"}
+		user2 := models.User{"Evan"}
```
ethomas2, line 80: 😮

handlers/handlers.go
atmiguel: To be clear we're getting the users' names, not the usernames (we don't currently have "usernames"). But we're getting them so that we can display the name of the user who made the comment

atmiguel: oh if you're asking why do we get the names of ALL of our users, that was just quicker. I was imagining a more refined url with query params, but it didn't seem necessary at this point. And we should talk about what that would look like.

handlers/handlers.go
atmiguel: what do you mean by non web-page api?

models/note.go
atmiguel: oh yeah ample comments would probably make sense here. It converts it into a byte array bc that's just what the builtin json.Marshall (equivalent to json.dumps in python) returns. The Alias thing was something i found online for custom json parsing. The reason this needed custom parsing was because we need to convert NoteType from being an integer into some readable form for the front end

static/js/login_or_signup.js
atmiguel: i'm sure there is a way to do it, though it may have to be kinda hacky. We ran into this issue at bevspot and we did some silly shit to get it working the way we wanted

atmiguel: definitely something we can do, just not a priority

static/js/notes.js
atmiguel: Yeah I can see how that can be confusing. I made a rule with myself that if a function returns a jQuery object it starts with a `$` which makes saying DOM or HTML at the end of the function somewhat redundant

static/js/notes.js
atmiguel: database side? i assume you just mean server side? We could certainly change the notes api to get the names and fill them in before giving it to the client side, but that seemed like it was mucking up the purity of the notes api.

handlers/handlers.go
atmiguel: 🤐

handlers/handlers.go
zingales: I consider non web-page as something that's doesn't return a html DOM object like `<html><body><div>lama</div></body></html>`

models/note.go
zingales: Not sure i understand completely. But it somehow helps to make a custom json serializer?

static/js/notes.js
zingales: I agree, but i'm not sure why. Would you mind explaining how it mucks with the purity of the notes api?

static/js/notes.js
zingales: I see so everything that starts with $ will always return pure html. So i could just read it as HtmlCreateAuthor.

zingales (approved):
Approved though still interested in answering my questions.

handlers/handlers.go
atmiguel: the convention is for them to return whatever data is being asked for or a status code explaining why it couldn't get the data. So in this case, we really should just delete this line

models/note.go
atmiguel: yup. anything i can clarify?

static/js/notes.js
atmiguel: not pure html, but a jQuery object

static/js/notes.js
atmiguel: i suppose it's just making it less normalized. So yeah for example if two people had the same name, there would be no way to tell who wrote the note. And you could imagine showing more user data in the future, like showing a picture next to the name like in github. You wouldn't want to then have to update the note api to also pass down that data.

static/js/notes.js
zingales: This isn't a big deal, but it helps me understand conventions. If the function didn't start with $ and it created a pure html object we'd call it something HtmlCreateAuthor?

models/note.go
zingales: Would it still have worked if you didn't create the Alias type? aka find and replaced Alias with Note and removed line 43?  Or is it used so that Golang specifically doesn't use the Type Note marshaling?

atmiguel: not sure. I expect converting note to type Note directly instead of indirectly might be problematic? But I really don't know. Can't say I entire understand what I've done. Perhaps we should see if @ethomas2 can explain what going on here

static/js/notes.js
atmiguel: I would probably call that `createAuthorElement` or if I knew more specifically what it was I might say that. For example `createAuthorDiv` or `createAuthorListItem`.

models/note.go
ethomas2: You can just nuke this whole method.
If you just call [json.Marshall()](https://golang.org/pkg/encoding/json/#Marshal) or [json.NewDecoder()](https://golang.org/pkg/encoding/json/#NewDecoder) on any arbitrary struct, it will use reflection to figure out the name of your fields (and act recursively on any values which are structs themselves). So if you nuke this method you'll get a blob like
```
{
  "AuthorId": <authorid>,
  "Type": <type>,
  "Content": <content>,
  "PublicationId": <publicationid>,
  "CreationTime": <creationtime>
}
```
If you want to change the name of those keys to make them lowercase or even change the names completely you can add ``` `json:<name>` ``` in the definition of the struct (you've already done that on lines 35-40. It's making the names of the fields lowercase, but you can use that comment to change it to any arbitrary name).

By defining your own MarshallJSON function on note, you've made Note implement the [marshaller interface](https://golang.org/pkg/encoding/json/#Marshaler). By making Note a Marshaller you're overriding the default reflection-based behavior of `json.Marshall`. Now,  instead of using reflection,`json.Marshall` will just call this custom function that you've defined and it will render a Note to whatever arbitrary thing you return out of this function. But the arbitrary thing you're returning out of this function is the same as what `json.Marshall` would return on it's own using reflection, so it's useless.

I _think_ the reason Alias is useful here is to avoid infinite recursion. Suppose you wanted to keep this method around for some reason and you wrote it like this
```go
function (note Note) MarshalJSON() (...) {
  return json.Marshall(&struct{
    Type: string
    Note
  }{
    Type: type,
    Note: note
  })
}
```
what would happen is you would call `json.Marshall(note)`, which would call `MarshallJSON`, which would then need to figure out how to render a Note on the third to last line. So it would call `json.Marshall(note)` and recurse infinitely.

That's a theory, you'd probably want to test it to confirm that. Also, it's weird that the struct that you have on lines 45-48 has Type at the top level and also Type in Alias (which is an embedded struct). I'm surprised the complier doesn't complain here.

zingales: Sorry i'm not sure i follow how you would get the side affect of line 49, while getting ride of this method. We added this because it was inserting the Type number instead of calling the String method.

PR #25, opened 2018-06-18
fixed compilation error

zingales:


models/note.go
```diff
@@ -16,7 +16,7 @@ const (
 	PREDICTIONS
 )
 
-const noteTypeStrings = [...]string{
+var noteTypeStrings = [...]string{
```
atmiguel, line 19: pls explain

atmiguel (approved):

models/note.go
zingales: i get a compilation when its const: `models/note.go:19:7: const initializer [4]string literal is not a constant`. Google tells me its cause you can't make arrays const, since they are mutable by nature.  
https://stackoverflow.com/a/13140094

PR #26, opened 2018-07-22
Cleaned up database functions

atmiguel:
As well as making a change to a handler function that evan recommended

### DIFF
--- a/databaseutil/databaseutil.go
+++ b/databaseutil/databaseutil.go
@@ -18,6 +18,12 @@ var db *sql.DB
 // UniqueConstraintError is returned when a uniqueness constraint is violated during an insert.
 var UniqueConstraintError = errors.New("postgres: unique constraint violation")
 
+// QueryResultContainedMultipleRowsError is returned when a query unexpectedly returns more than one row.
+var QueryResultContainedMultipleRowsError = errors.New("query result unexpectedly contained multiple rows")
+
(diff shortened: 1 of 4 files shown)


databaseutil/databaseutil.go
```diff
@@ -60,36 +66,58 @@ func InsertIntoUsersTable(
...
+	rows, err := db.Query(sqlQuery, emailAddress)
+	if err != nil {
+		return 0, convertPostgresError(err)
 	}
+	defer rows.Close()
 
 	var userId int64
-	if err := row.Scan(&userId); err != nil {
-		return 0, err
+	for rows.Next() {
```
zingales, line 109: The next time we do this exact logic we should make it its own separate function.

handlers/handlers.go
```diff
@@ -334,15 +333,14 @@ func HandleNotesPageRequest(
 
 func respondWithMethodNotAllowed(
 	responseWriter http.ResponseWriter,
-	allowedMethods ...string,
+	allowedMethod string,
```
zingales, line 336: this is to guarantee there is at least one method, at compile time?

atmiguel: yeah

databaseutil/databaseutil.go
atmiguel: agreed

zingales (approved):

zingales (approved):

PR #28, opened 2018-08-01
[Work in Progress] Add note backend

zingales:


databaseutil/databaseutil.go
```diff
@@ -42,7 +43,7 @@ func ConnectToDatabase(databaseUrl string) error {
 	return nil
 }
 
-func InsertIntoUsersTable(
+func InsertIntoUserTable(
```
atmiguel, line 46: hmm, so it seems like we're trying to be consistent in our use of singular/plural, but now our database tables don't have the same name as this function implies

databaseutil/databaseutil.go
```diff
@@ -65,6 +66,48 @@ func InsertIntoUsersTable(
...
+	content string,
+	publicationId int64,
+	creationTime time.Time,
+) error {
+
+	sqlQuery := `
+		INSERT INTO notes (author_id, type, content, publication_id, creation_time)
+		VALUES ($1, $2, $3, $4, $5)`
+
+
```
atmiguel: remove extra newline

atmiguel: are you not using gofmt?

databaseutil/databaseutil.go
```diff
@@ -65,6 +66,48 @@ func InsertIntoUsersTable(
...
+	var cleanPublicationId sql.NullInt64
+	var cleanNoteType sql.NullString
+
+	if publicationId < 1 {
```
atmiguel: Wouldn't it make more sense for `publicationId` to be a nullable int? Instead of the caller having to remember that to indicate a null publicationId they have to specify a value less than 1?

databaseutil/databaseutil.go
```diff
@@ -65,6 +66,48 @@ func InsertIntoUsersTable(
 	return nil
 }
 
+
+func InsertIntoNoteTable(
+	userId int64,
+	noteType string,
```
atmiguel: i forget, did we make the convention that this file can't have any types specific to our application logic? Is that why this type is string and not NoteType?

databaseutil/databaseutil.go
```diff
@@ -65,6 +66,48 @@ func InsertIntoUsersTable(
...
+	var cleanNoteType sql.NullString
+
+	if publicationId < 1 {
+		cleanPublicationId = sql.NullInt64{Int64:0, Valid:false}
+	} else {
+		cleanPublicationId = sql.NullInt64{Int64:publicationId, Valid:true}
+	}
+
+	if len(noteType) == 0 {
+		cleanNoteType = sql.NullString{String:"", Valid:false}
```
atmiguel: `noteType` should be nullable

atmiguel: alternatively, we could make other functions that didn't take noteType as a parameter and just shared logic with this function internally

databaseutil/databaseutil.go
```diff
@@ -65,6 +66,48 @@ func InsertIntoUsersTable(
...
+	} else {
+		cleanNoteType = sql.NullString{String: strings.ToLower(noteType), Valid:true}
```
atmiguel: `noteType` should already be in exactly the form postgres expects by the time we get here, so there should be no need to lowercase it

handlers/handlers.go
```diff
@@ -251,8 +251,38 @@ func HandleNoteApiRequest(
...
+	case http.MethodPost:
+		type NoteForm struct {
+			Content  string `json:"content"`
+			NoteType string `json:"noteType"`
+		}
+
+		noteForm := new(NoteForm)
+
+		if err := json.NewDecoder(request.Body).Decode(noteForm); err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
```
atmiguel: This shouldn't be a 500. This should only occur when the client sends invalid/unexpected data

handlers/handlers.go
```diff
@@ -251,8 +251,38 @@ func HandleNoteApiRequest(
...
+			return
+		}
+		var statusCode int
```
atmiguel: looks like this is only ever assigned once

handlers/handlers.go
```diff
@@ -251,8 +251,38 @@ func HandleNoteApiRequest(
...
+		if len(strings.TrimSpace(noteForm.Content)) == 0 {
+			http.Error(responseWriter, "Note content cannot be empty or just whitespace", http.StatusBadRequest)
+			return
+		}
+		note := models.CreateNewNote(userId, noteForm.Content, models.DecodeNoteType(noteForm.NoteType))
+
+		if err := userservice.StoreNewNote(note); err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+			return
+		} else {
```
atmiguel: the if block above returns, so there's no need for an else block

models/note.go
```diff
@@ -23,14 +24,37 @@ var noteTypeStrings = [...]string{
 	"Predictions",
 }
 
+func DecodeNoteType(input string) NoteType {
```
atmiguel: how about `DeserializeNoteType`?

models/note.go
```diff
@@ -23,14 +24,37 @@ var noteTypeStrings = [...]string{
...
+	for i := 0; i < len(noteTypeStrings); i++ {
+		if strings.ToLower(input) == strings.ToLower(noteTypeStrings[i]) {
```
atmiguel: we control the capitalization of both of these things. We shouldn't have to lowercase them to be sure they are equal

models/note.go
```diff
@@ -23,14 +24,37 @@ var noteTypeStrings = [...]string{
...
+			return NoteType(i);
+		}
+	}
+	return NoteType(-1);
```
atmiguel: what does this do?

atmiguel: shouldn't we throw an exception?

models/note.go
```diff
@@ -23,14 +24,37 @@ var noteTypeStrings = [...]string{
...
+}
+
 func (noteType NoteType) String() string {
 	if noteType < MARGINALIA || noteType > PREDICTIONS {
-		return "Unknown"
+		return ""
```
atmiguel: this should probably be an exception as well

models/note.go
```diff
@@ -23,14 +24,37 @@ var noteTypeStrings = [...]string{
...
 	}
 
 	return noteTypeStrings[noteType]
 }
 
+
+
+
+func CreateNewNote(userId UserId, content string, noteType NoteType) *Note {
```
atmiguel: why do we need this function?

models/note.go
```diff
@@ -23,14 +24,37 @@ var noteTypeStrings = [...]string{
...
+	note := new(Note);
```
atmiguel: can't this be done in one line?

models/note.go
```diff
@@ -23,14 +24,37 @@ var noteTypeStrings = [...]string{
...
+	note.AuthorId = userId;
+	note.Content = content;
+	note.Type = noteType;
```
atmiguel: where did semicolons come from?? pls use gofmt

services/userservice/userservice.go
```diff
@@ -46,6 +46,20 @@ func StoreNewUser(
 	return nil
 }
 
+func StoreNewNote(note *models.Note) error {
```
atmiguel: this doesn't belong in userservice

services/userservice/userservice.go
```diff
@@ -46,6 +46,20 @@ func StoreNewUser(
...
+	if err:= databaseutil.InsertIntoNoteTable(
+		int64(note.AuthorId),
```
atmiguel: is there another way to get the value? like `.value()` or something? Just seems weird that we would have to cast it

templates/notes.tmpl
```diff
@@ -15,5 +15,28 @@
...
+        <div id="new-note"> 
+        	<h3> New Note </h3>
+			<div class="mui-form">
+				<form>
+        			<h4> New Note </h4>
+					<input type="radio" name="note_type" value="not_categorized" checked="checked"> Undecided<br>
+                    <input type="radio" name="note_type" value="predictions"> Predictions<br>
+					<input type="radio" name="note_type" value="marginalia">Marginalia<br>
+    				<input type="radio" name="note_type" value="meta">Meta<br>
+    				<input type="radio" name="note_type" value="questions">Qustions<br>
```
atmiguel: have you no respect for indentation?

templates/notes.tmpl
```diff
@@ -15,5 +15,28 @@
...
+                <br />
+                 <div class="mui-textfield mui-textfield--float-label">
+                    <textarea> </textarea>
+                    <label>Content</label>
+                    <span class="mui--text-accent-secondary"></span>
+                </div>
+                <button type="button" class="mui-btn mui-btn--primary">
+                	Save Note
+				</button>
+             </div>
```
atmiguel: also, i think i already made this? did you not see it? or did i not make a PR?

atmiguel (changes requested):

databaseutil/databaseutil.go
zingales: So go doesn't have a builtin nullable int type. 

options and why i didn't choose them
1) publicationId should be a nullable int in the note.go file? 
If so i don't like that option since nullable int a a sql exclusive object, and i wanted to keep our datasystem storage agnostic. 

2) why doesn't this function just accept a nullabelint instead of an int. Because you can't just assign pass in a int to a sql.NullableInt you have to create the struct you see in the next line. So that would mean the calling function would need to create the struct which i don't like.

databaseutil/databaseutil.go
zingales: That's what i remember. Databaseutil is application agnostic, and user service is storage agnostic. 

Though that rule is not written in stone.

databaseutil/databaseutil.go
zingales: When you say noteType should be nullable what do you mean? 


1) go doesn't allow you to assign nil
2) add an element to the struct which is equivalent to null. Like unassigned?

models/note.go
zingales: Didn't realize that it had been turned off. Turned it back on.

services/userservice/userservice.go
zingales: You think it should go in databaseutil? This is where i handle the interface of application contrext to database context.

templates/notes.tmpl
zingales: i didn't see it. Thank god you made it. I was struggling pretty bad.

services/userservice/userservice.go
zingales: there is no .value(). But i think for the compiler we'll need to cast it. The compiler things authorid and int64 are different types. So there needs to be a function that translates it. If you really feel weird, we can add a function to UserId which just casts it behind the scenes and call it value. But that seems a bit circuitous

databaseutil/databaseutil.go
zingales: I can change it to users. I didn't remember that the table was called users. Though i fear we are quickly approaching the ROR conundrum. Are all tables plural versions of thier row element. Luckily its not a big deal to rename the tables.

models/note.go
zingales: we'll need to discuss how to handle the "nullable" note type.

services/userservice/userservice.go
zingales: You know what i change my mind. database util with be the interface between application and storage. 

Though this make me reconsider the purpose of user service.

databaseutil/databaseutil.go
atmiguel: yeah i think i agree. fuck plurals. let's just make a task to rename the tables to singular

databaseutil/databaseutil.go
atmiguel: oh okay i didn't realize we couldn't do nullable int. I thought golang allowed something like:
```
var nullableNumber : int | null
```
or something like that, but i think i'm just confused

atmiguel: So i agree our note struct shouldn't necessarily depend on a sql specific struct, but it should also be an accurate representation of our model. A 0 or -1 for publicationId is just not the same thing as null. One option would be to create our own concept of nullableInt

atmiguel: Another option that I've been thinking about (but would involve a larger discussion) is to make our database rows immutable. Meaning notes wouldn't have publicationId on them, but there would be a separate association table relating noteIds to publicationIds

atmiguel: this would kind of avoid the problem altogether. we would retrieve notes with all non-nullable fields, and we would retrieve a separate thing that would look like Map<noteId, publicationId>

databaseutil/databaseutil.go
atmiguel: hmm yeah seems fine for now, but may need more thought in the future

databaseutil/databaseutil.go
atmiguel: wut string isn't nullable?

atmiguel: jesus, this is reaffirming my immutable db rows thing. instead of having nullable noteTypes we could just have a separate association table

services/userservice/userservice.go
atmiguel: no sorry i should've been more clear. i was trying to point out that this function has nothing to do with "users" so it doesn't belong in userservice. There should be a new file called noteservice

services/userservice/userservice.go
atmiguel: christ

atmiguel: Calling all evans. Is this the way this is expected to be done? I thought casting was to be avoided whenever possible. Is there compiler-level assurance that this won't ruin everything?

handlers/handlers.go
```diff
@@ -260,26 +259,24 @@ func HandleNoteApiRequest(
...
 
 		if err := userservice.StoreNewNote(note); err != nil {
 			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
 			return
-		} else {
-			statusCode = http.StatusCreated
 		}
 
-		responseWriter.WriteHeader(statusCode)
+		statusCode := http.StatusCreated
```
atmiguel: no need to do this in two lines

models/note.go
```diff
@@ -24,13 +24,13 @@ var noteTypeStrings = [...]string{
 	"Predictions",
 }
 
-func DecodeNoteType(input string) NoteType {
+func DeserializeNoteType(input string) NoteType {
 	for i := 0; i < len(noteTypeStrings); i++ {
 		if strings.ToLower(input) == strings.ToLower(noteTypeStrings[i]) {
```
atmiguel: bump

models/note.go
```diff
@@ -41,14 +41,11 @@ func (noteType NoteType) String() string {
 	return noteTypeStrings[noteType]
 }
 
-
-
-
 func CreateNewNote(userId UserId, content string, noteType NoteType) *Note {
```
atmiguel: bump

databaseutil/databaseutil.go
zingales: I agree with having a reference table, <NoteId, PublicationId> Table, and not storing publicationId in the table! Lets do that. Want me to do that as part of this change? or make a seperate pull request for that agree on that. Then come back to this?

services/userservice/userservice.go
zingales: That makes sense! You want a new package called note service?

models/note.go
```diff
@@ -2,51 +2,81 @@ package models
...
+	"predictions",
+}
+
+func DeserializeNoteType(input string) (NoteType, error) {
+	for i := 0; i < len(noteTypeStrings); i++ {
+		if strings.ToLower(input) == strings.ToLower(noteTypeStrings[i]) {
+			return NoteType(i), nil
+		}
+	}
+	return UNCATEGORIZED, UnDeserializeableNoteTypeStringError
```
zingales, line 40: I hate putting a value here like uncategorized, however, the go compiler wouldn't allow me to put nil.

PR #29, opened 2018-08-02
[Work In Progress] Add note front-end

atmiguel:


static/js/notes.js
```diff
@@ -38,7 +100,30 @@ const $createNote = function(note) {
...
+            .append($createRowWithTwoElements($buttons[2], $buttons[3])))
+        .append($textarea);
+};
+
+const activateModal = function($modal) {
+    mui.overlay('on', $modal.get(0));
+};
+
 $(function() {
+    const $addNoteModal = $createAddNoteModal();
```
zingales, line 125: I was pretty confused by this for a while, i was like its a function that adds a note modal. But i was like its seems like its a static object, its not a function that will add a note modal. 

Then it finally hit me its a modal to add notes. *facepalm*

I know we've called it add note everywhere. But maybe we should call it New Note and new Note modal?

static/js/notes.js
```diff
@@ -52,4 +137,26 @@ $(function() {
             });
         });
     });
+
+    $('#add-note-button').click(function() {
+        activateModal($addNoteModal);
+    });
+
+    $(document).on('click', classesByName.noteTypeButton, function() {
```
zingales, line 145: is this to allow you to get radio button like features. Where in clicking one note type it de-selects any other notetype that was selected? 

And we created a new one from scratch because radio buttons don't allow you to deselect/ leave it blank?

static/js/notes.js
```diff
@@ -38,7 +100,30 @@ const $createNote = function(note) {
...
+
+    return $modal
+        .append($createGridContainer()
+            .append($createRowWithTwoElements($buttons[0], $buttons[1]))
+            .append($createRowWithTwoElements($buttons[2], $buttons[3])))
+        .append($textarea);
+};
+
+const activateModal = function($modal) {
+    mui.overlay('on', $modal.get(0));
```
zingales, line 121: If i wanted to rename this code-ugliness what would you recommend we call it?  `$obj.getRawDomObject()`?

zingales:
This isn't caused by this code change, but i was wondering if this code commit could also handle this: 

screen shot 2018-08-27 at 1 40 13 pm

Essentially the new note object that we return doesn't include a note category. So when its empty it just looks like 2 dashes since it inserts an empty string.

static/js/notes.js
zingales: I'm leaning heavily towards adding our own function to the jquery prototype to do this.

atmiguel: Can you explain what you mean? What specifically do you find ugly about this?

static/js/notes.js
atmiguel: Yeah maybe. I don't think add is a common verb for putting something in the DOM

static/js/notes.js
atmiguel: That's was part of the reason, yes. The other reason being this was the UI we had agreed to I think. We designed for mobile first.

zingales: I'm totally cool with this. Just to know on the side, are radio buttons bad on mobile?

static/js/notes.js
zingales: Sorry for my vagueness. I dislike `$modal.get(0)`. I find it very unreadable, and counter intuitive. How do you know 0 is defined? What else is in the list? On top of that everytime i see it, my first thought is that the `$obj` is poorly named because i don't know what the `.get(0)` does. Then i remember that its this hack to get the `rawDomHtml` and i'm like oh when i have some extra time i should write a class to fix that.

zingales: how does this look to you? https://github.com/atmiguel/cerealnotes/blob/editApiv2/static/js/base.js It means that in future code edits you can do `$modal.getDOM()`

static/js/notes.js
atmiguel: uh i'm honestly not sure. @riceballthief do you know?

static/js/notes.js
atmiguel: Why do you consider `.get()` a hack? `getDOM` is a bit confusing. Perhaps something like `getUnderlyingDOMElement`?

zingales: `getUnderlyingDomElement` sounds great. Please note that i specifically fixed my capitalization error and didn't capitalize `OM` in `DOM`. Its a convention at Microsoft which i now think is great. Simply put in camel case, everything is lowercase except for the first letter. This helps read the names clearly especially when there are multiple acronyms involved. Imaging you had a convention in your code to have ON mean Oriented North. Now this method is ambiguous `getJSONPut`. is it supposed to be `get JSON Put` or `get JS ON Put`? However, with the only one capital rule its clear. it'd be `getJsOnPut` or `getJsonPut` (granted you'd never want to have both of these methods in your code)

static/js/notes.js
riceballthief: Radio buttons are only use when there are two ore more mutually exclusive options and the users must select only one choice. It's usually good practice to use them to change settings but  never as action buttons. Also, in terms of visual design, it's usually best to arrange items vertically in a logical order. It's not bad to use on mobile in general, there're just appropriate and inappropriate situations to use them, and this goes for non-mobile as well.

PR #30, opened 2018-08-06
Database Refactoring Option A

zingales:


models/note.go
```diff
@@ -10,33 +10,34 @@ type NoteId int64
...
 	PREDICTIONS
 )
 
 var noteTypeStrings = [...]string{
-	"Marginalia",
-	"Meta",
-	"Questions",
-	"Predictions",
+	"uncategorized",
+	"marginalia",
```
zingales, line 22: I lower all these strings so they would match 1:1 with the strings in the Databases enum. Since when i compared them in the past the DB was case sensitive for enum matching.

models/note.go
```diff
@@ -10,33 +10,34 @@ type NoteId int64
 type NoteType int
 
 const (
-	MARGINALIA NoteType = iota
+	UNCATEGORIZED NoteType = iota
```
zingales, line 13: I felt like uncategorized label, should either be at the begging or at the end of the enum, since in some ways it stands apart from the other categories. Putting it at the end would cause its "number" to change as we added more categories.

migrations/0000_createDbs.sql
```diff
@@ -1,26 +1,31 @@
 -- Types
-CREATE TYPE note_type AS ENUM ('predictions', 'marginalia', 'meta', 'questions');
+CREATE TYPE note_type AS ENUM ('uncategorized', 'predictions', 'marginalia', 'meta', 'questions');
```
atmiguel, line 2: I think it might make more sense to have another relation table for note_type instead of adding this category

migrations/0000_createDbs.sql
```diff
@@ -1,26 +1,31 @@
...
 	type note_type,
 	content text NOT NULL,
-	publication_id bigint references publications(id),
 	creation_time timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS publication_to_note_linkage(
+	id bigserial PRIMARY KEY,
+	publication_id bigint references publication(id) NOT NULL,
+	note_id bigint references note(id) NOT NULL
```
atmiguel: `note_id` should be `UNIQUE`

migrations/0000_createDbs.sql
```diff
@@ -1,26 +1,31 @@
...
 	id bigserial PRIMARY KEY,
-	author_id bigint references users(id) NOT NULL,
+	author_id bigint references user(id) NOT NULL,
 	type note_type,
 	content text NOT NULL,
-	publication_id bigint references publications(id),
 	creation_time timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS publication_to_note_linkage(
```
atmiguel: `linkage` sounds weird to me, but i'm pretty sure it's not wrong. alternatives would be appreciated, but unnecessary

atmiguel (changes requested):

migrations/0000_createDbs.sql
zingales: Having a table that relates note_type and note seems like it assumes the wrong things. The one thing i like about it being in the row itself is that its clear that a note can only have one type. Which i know you can also do with sql restrictions, but it feels cleaner. 

I do understand that to keep your vision pure about making notes immutable, it would have to be referenced since this is an allowed change. However, i'm not i fully agree with that vision. At least not right now. It might be something that we want, but i think we should cross that bridge when we get there and only make it a requirement now if we know we will add that feature later.

migrations/0000_createDbs.sql
zingales: I agree, i'm not a fan of this table name either. I just couldn't think of any better alternatives. ptn_reference seems to vague, ptn_association also vague. Maybe ptn_relationship?

migrations/0000_createDbs.sql
zingales: good call

migrations/0000_createDbs.sql
atmiguel: I'm not sure how you can have those reservations about noteType but not about publication. Why should they be treated differently?

migrations/0000_createDbs.sql
atmiguel: relationship might be a bit better. it's all pretty vague, but maybe that's okay

migrations/0000_createDbs.sql
zingales: In my mind tables tend to be of two types, relationship tables and entity tables. Each entity table contains an "object" which is standalone and useful to talk about on its own. 

You can also have relationship tables, which define how entity tables are related, sometimes they have extra columns that define different characteristics of the relationship. 

So back to the matter at hand, for me publications and notes are very clearly two entities, and so i have no issue with having two tables for them. These entities have a relationship, which is 1:many and it makes sense to have a relationship table. 

The issue i have with note type is that its a descriptor not a separate object. So a table that defines a linkage between a notetype and a note doesn't fit into my simple world of entity and relationship tables. It feels like an addendum table which defines more characteristics of the note, rather than a relationship to a between objects. 

Granted I understand that this is _my_ table vision and so by no means the golden standard. But you were asking why i saw them differently.

PR #31, opened 2018-08-22
Database Refactoring Option B

zingales:
This is option b for the db refactor.


migrations/0000_createDbs.sql
```diff
@@ -2,25 +2,35 @@
...
 	email_address text UNIQUE NOT NULL,
 	password bytea NOT NULL,
 	creation_time timestamp NOT NULL
 );
 
-CREATE TABLE IF NOT EXISTS publications(
+CREATE TABLE IF NOT EXISTS publication (
 	id bigserial PRIMARY KEY,
-	author_id bigint references users(id) NOT NULL,
+	author_id bigint references app_user(id) NOT NULL,
```
atmiguel: just remembering that we need to have some sort of foreign key assurance that all note to publication relationships refer to the same author

migrations/0000_createDbs.sql
```diff
@@ -2,25 +2,35 @@
...
 	content text NOT NULL,
-	publication_id bigint references publications(id),
 	creation_time timestamp NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS publication_to_note_relationship (
+	id bigserial PRIMARY KEY,
+	publication_id bigint references publication(id) NOT NULL,
+	note_id bigint UNIQUE references note(id) NOT NULL
```
atmiguel: since this is unique and not null, we can just make this column the primary key and remove the `id` column

migrations/0000_createDbs.sql
```diff
@@ -2,25 +2,35 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS notetype_to_note_relationship (
```
atmiguel: how about `note_to_type_relationship`?

atmiguel: or perhaps just adding an underscore to your current name works better:
`notetype...` --> `note_type...`

migrations/0000_createDbs.sql
```diff
@@ -2,25 +2,35 @@
...
+	id bigserial PRIMARY KEY,
+	type note_type,
```
atmiguel: this should not be nullable

migrations/0000_createDbs.sql
```diff
@@ -2,25 +2,35 @@
...
+	note_id bigint UNIQUE references note(id) NOT NULL
```
atmiguel: since this is unique and not null, this could just be the primary key

migrations/drop_everything._sql
```diff
@@ -0,0 +1,11 @@
+DROP TYPE note_type CASCADE;
+
```
atmiguel: what's with the file name? or specifically the extension?

atmiguel (changes requested):

migrations/drop_everything._sql
zingales: Its so you never accidentialy run it. For example i could imagine someone (e.g. Evan) run all migrations that end in .sql. This would accidentally get run. If you'd like i could change it to .txt. But i thought or .sql.txt.

migrations/0000_createDbs.sql
zingales: What's the difference of having a note with notetype nullable in the relationship table? Wouldn't it be more straightforward to just not have a row for that note?

migrations/0000_createDbs.sql
zingales: how do we do that?

atmiguel: I'm honestly not sure of the best way, but I can think of _a_ way to do it

atmiguel: If we add an author_id column to the publication_to_note_relationship table, then make two separate foreign keys on that table, that will work, but it's a bit ugly

atmiguel (approved):

migrations/drop_everything._sql
atmiguel: another option might be to put it in a separate folder? or a subfolder?

PR #32, opened 2018-08-27
Rename note_type to cagtegory

zingales:


migrations/0000_createDbs.sql
```diff
@@ -29,7 +29,7 @@ CREATE TABLE IF NOT EXISTS note_to_publication_relationship (
 	publication_id bigint references publication(id) NOT NULL
 );
 
-CREATE TABLE IF NOT EXISTS note_to_type_relationship (
+CREATE TABLE IF NOT EXISTS note_to_category_relationship (
 	note_id bigint PRIMARY KEY references note(id),
-	type note_type NOT NULL
+	type category NOT NULL
```
atmiguel: the column name shouldn't be `type`. It feels like it should be `category`, but that is clearly problematic. What about renaming the type to `category_type` or `note_category`?

atmiguel (changes requested):

routers/routers.go
```diff
@@ -10,9 +10,20 @@ import (
 	"github.com/atmiguel/cerealnotes/paths"
 )
 
+type routeHandler struct {
```
zingales, line 13: This "embeds" a type which is super cool. It allows callers to access the methods of the embedded type without need endless wrapper methods.

routers/routers.go
```diff
@@ -27,28 +45,21 @@ func DefineRoutes() http.Handler {
 
 	// Redirects
 	mux.HandleFunc("/", handlers.RedirectToPathHandler(paths.HomePage))
+	mux.HandleFunc("/api/", http.NotFound)
```
zingales, line 48: without this line. /api/bobby will redirect to /home. Which can be confusing if your api calls returns a redirect, then resolves to the home page. Now for bad invalid api calls it just returns 404.

routers/routers.go
```diff
@@ -10,9 +10,27 @@ import (
 	"github.com/atmiguel/cerealnotes/paths"
 )
 
+type routeHandler struct {
+	*http.ServeMux
+}
+
+func (mux *routeHandler) handleAuthenticated(
+	pattern string,
+	handlerFunc handlers.AuthentictedRequestHandlerType,
```
atmiguel: `AuthentictedRequestHandlerType` has a typo. It's missing the second `a` in Authentic**a**ted

routers/routers.go
atmiguel: So it's almost like adding a function to a prototype in javascript?

routers/routers.go
```diff
@@ -10,9 +10,27 @@ import (
 	"github.com/atmiguel/cerealnotes/paths"
 )
 
+type routeHandler struct {
+	*http.ServeMux
+}
+
+func (mux *routeHandler) handleAuthenticated(
```
atmiguel: with the other function now in scope, perhaps this should be `handleAuthenticatedPage`?

routers/routers.go
atmiguel: I'm a bit confused, mostly because i forgot how go matches against these routes we pass in. If someone tries to go to `/random` right now, does it return a 404 or a redirect?

handlers/handlers.go
```diff
@@ -256,17 +256,36 @@ type AuthentictedRequestHandlerType func(
...
+			default:
+				respondWithMethodNotAllowed(responseWriter, http.MethodGet)
+			}
+		} else {
+			authenticatedHandlerFunc(responseWriter, request, userId)
+		}
+	}
+}
+
+func AuthenticateOrUnauthorized(
```
atmiguel: This name is a bit confusing. I typically expect functions to be actions/verbs and if there's an `Or` statement I'd expect `VerbOrVerb`. I can't tell what this function does by reading the name

handlers/handlers.go
```diff
@@ -256,17 +256,36 @@ type AuthentictedRequestHandlerType func(
...
+	authenticatedHandlerFunc AuthentictedRequestHandlerType,
+) http.HandlerFunc {
+	return func(responseWriter http.ResponseWriter, request *http.Request) {
+		if userId, err := getUserIdFromJwtToken(request); err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusUnauthorized)
```
atmiguel: Found this on stack overflow (https://stackoverflow.com/a/6937030):

>There's a problem with 401 Unauthorized, the HTTP status code for authentication errors. And that’s just it: it’s for authentication, not authorization. Receiving a 401 response is the server telling you, “you aren’t authenticated–either not authenticated at all or authenticated incorrectly–but please reauthenticate and try again.” **To help you out, it will always include a WWW-Authenticate header that describes how to authenticate.**

Seems like the correct response. We just need to send back an additional header

migrations/0000_createDbs.sql
```diff
@@ -29,7 +29,7 @@ CREATE TABLE IF NOT EXISTS note_to_publication_relationship (
 	publication_id bigint references publication(id) NOT NULL
 );
 
-CREATE TABLE IF NOT EXISTS note_to_type_relationship (
+CREATE TABLE IF NOT EXISTS note_to_category_relationship (
 	note_id bigint PRIMARY KEY references note(id),
-	type note_type NOT NULL
+	category_type category NOT NULL
```
atmiguel: sorry if my last comment wasn't clear, but i was suggesting the opposite change. I would expect the `_type` to go on the type, not on the column name

atmiguel (changes requested):

routers/routers.go
zingales: Page that doesn't exist: 
/invalid -> redirect to /home
/api/invalid -> 404

not logged in:
/unauthorized -> redirect to /home
/api/unauthorized -> 401

handlers/handlers.go
zingales: how about `AuthenticateOrReturnUnauthorized` a bit verbose but clear no?

migrations/0000_createDbs.sql
zingales: That makes a lot more sense.

routers/routers.go
ethomas2: I consider it to be more like classical inheritance than prototype inheritance. Type embedding _is_ like prototype inheritance, but only because prototype inheritance is like classical inheritance. Off the top of my head I can't think of anything similar between type embedding and prototype inheritance that isn't also similar between type embedding and classical inheritance, so comparing it to prototype inheritance is a bit misleading.

For example, with prototype inheritance you can dynamically add/remove functions to the prototype at runtime and all objects will be affected, even objects that were instantiated prior to messing with the prototype. Type embedding doesn't have anything like that

routers/routers.go
atmiguel: hmm okay. I guess I'm confused about what we're going for? I feel like there's a lot of different behaviors here. Is there a reason we don't just return a 404 for all endpoints we have yet to define? Wouldn't that be the most correct?

handlers/handlers.go
atmiguel: I don't particularly like having the verb return in a function, but given that i have no alternatives and the function name isn't wrong let's go with it for now

atmiguel: I think I just realized what pattern I'd like to end up at eventually. Typically an issue with naming a function is an issue the function was poorly designed. This function is doing two things. Ideally it would just authenticate, and return an error if it failed to do so. We did this as a convenience to not have to put in error handling that returns 401.

I totally like the idea, but I think what we want to do eventually is have some wrapper around all of our handlers that just checks the error that was returned and translates that to an http response code. For example, this authenticate function would return an error like `FailedToAuthenicateUserError` or `UserIsUnauthenticatedError`, and another function might return something like `MalformedPostBodyContentError`. And then the wrapper function would look at the error type and say okay, you're a 401, you're a 400, etc.

routers/routers.go
zingales: We could if you wanted to. The redirect all bad pages was a side-effect (i was pleasantly surprised with ) of adding a redirect for just /.

zingales: we could. I personally like the redirect for bad pages, but i could see it being frustrating for users.  

I also like having the control of behavior between api endpoints and web pages. The not-authenticated is a really good example. Before if you tried an api call and you forgot to included your cookie you'd get a temporary redirect to the login page. Now it warns you that you that you need a better bearer, including a message of where it failed in the process.

atmiguel (approved):

PR #33, opened 2018-08-27
Create api for adding notes to the database. 

zingales:


services/noteservice/noteservice.go
```diff
@@ -0,0 +1,47 @@
...
+func StoreNewNote(
+	note *models.Note,
+) error {
+
+	id, err := databaseutil.StoreNewNote(int64(note.AuthorId), note.Content, note.CreationTime)
+	if err != nil {
+		return err
+	}
+
+	note.Id = id
```
zingales: This line here makes me feel a bit better that we created a `<object>service` and `databaseutil` organization. Rather than just all of the stuff in one place.

databaseutil/databaseutil.go
```diff
@@ -94,6 +94,59 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
 	return password, nil
 }
 
+func StoreNewNote(authorId int64, content string, creationTime time.Time) (int64, error) {
```
atmiguel: It looks like the convention in this file is to use the verb `insert` instead of `store`

atmiguel: I'm realizing more and more that I think I made a mistake when I said we should only use builtin types in this file. I shouldn't be looking at that authorId wondering if someone passed in an authorId or not

databaseutil/databaseutil.go
```diff
@@ -94,6 +94,59 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
...
+	sqlQuery := `
+		INSERT INTO note (author_id, content, creation_time)
+		VALUES ($1, $2, $3)
+		RETURNING id`
+
+	rows, err := db.Query(sqlQuery, authorId, content, creationTime)
+	if err != nil {
+		return -1, convertPostgresError(err)
```
atmiguel: I believe our/go convention is to return 0, not -1

databaseutil/databaseutil.go
```diff
@@ -94,6 +94,59 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
...
+	}
+	defer rows.Close()
+
+	var lastInsertId int64
+	for rows.Next() {
+
+		if lastInsertId != 0 {
```
atmiguel: I like this, but can we explicitly initialize `lastInsertId` to `0`?

databaseutil/databaseutil.go
```diff
@@ -94,6 +94,59 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
...
+		VALUES ($1, $2, $3)
+		RETURNING id`
+
+	rows, err := db.Query(sqlQuery, authorId, content, creationTime)
+	if err != nil {
+		return -1, convertPostgresError(err)
+	}
+	defer rows.Close()
+
+	var lastInsertId int64
```
atmiguel: why not `noteId`?

databaseutil/databaseutil.go
```diff
@@ -94,6 +94,59 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
...
+	}
+
+	if err := rows.Err(); err != nil {
+		return -1, convertPostgresError(err)
+	}
+
+	return lastInsertId, nil
+}
+
+func StoreNoteCategoryRelationship(noteId int64, category string) error {
```
atmiguel: `Insert`

databaseutil/databaseutil.go
```diff
@@ -94,6 +94,59 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
...
+	sqlQuery := `
+		INSERT INTO note_to_category_relationship (note_id, type)
+		VALUES ($1, $2)`
+
+	rows, err := db.Query(sqlQuery, noteId, category)
+	if err != nil {
+		return convertPostgresError(err)
```
atmiguel, line 139: we need to explicitly handle the case where `note_id` already exists in the table

models/note.go
```diff
@@ -22,16 +23,33 @@ var categoryStrings = [...]string{
...
 	}
+	return MARGINALIA, UnDeserializeableCategoryStringError
+}
 
+func (category Category) String() string {
 	return categoryStrings[category]
 }
 
 type Note struct {
+	Id           int64     `json:"id"`
```
atmiguel: why are we adding id to note?

handlers/handlers.go
```diff
@@ -246,8 +249,51 @@ func HandleNoteApiRequest(
 
 		fmt.Fprint(responseWriter, string(notesInJson))
 
+	case http.MethodPost:
+		type NoteForm struct {
+			Content  string `json:"content"`
+			Category string `json:"category"`
+		}
```
atmiguel, line 255: I think these fields need to go in two separate api calls. What is category when none is selected?

handlers/handlers.go
```diff
@@ -246,8 +249,51 @@ func HandleNoteApiRequest(
...
+			http.Error(responseWriter, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		if len(strings.TrimSpace(noteForm.Content)) == 0 {
+			http.Error(responseWriter, "Note content cannot be empty or just whitespace", http.StatusBadRequest)
+			return
+		}
+
+		note := models.CreateNewNote(userId, noteForm.Content)
```
atmiguel: this confused me when i read it. Do we actually need this function? Why not just create the object here?

handlers/handlers.go
```diff
@@ -246,8 +249,51 @@ func HandleNoteApiRequest(
...
+
+		if err := noteservice.StoreNewNote(note); err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		if !(noteForm.Category == "") {
```
atmiguel: why not use `!=`?

handlers/handlers.go
```diff
@@ -246,8 +249,51 @@ func HandleNoteApiRequest(
...
+
+			category, err := models.DeserializeCategory(strings.ToLower(noteForm.Category))
```
atmiguel: we shouldn't have to use `ToLower`

handlers/handlers.go
```diff
@@ -246,8 +249,51 @@ func HandleNoteApiRequest(
...
+
+			if err != nil {
+				http.Error(responseWriter, err.Error(), http.StatusBadRequest)
+				return
+			}
+
+			if err := noteservice.StoreNoteCategoryRelationship(note, category); err != nil {
```
atmiguel: I think our convention is `StoreNewX`

models/note.go
```diff
@@ -22,16 +23,33 @@ var categoryStrings = [...]string{
 	"predictions",
 }
 
-func (category Category) String() string {
-	if category < MARGINALIA || category > PREDICTIONS {
-		return "Unknown"
+var UnDeserializeableCategoryStringError = errors.New("String does not correspond to a Note Category")
```
atmiguel: How about `CannotDeserializeCategoryStringError` or something like that

models/note.go
```diff
@@ -22,16 +23,33 @@ var categoryStrings = [...]string{
...
+	for i := 0; i < len(categoryStrings); i++ {
+		if input == categoryStrings[i] {
+			return Category(i), nil
+		}
 	}
+	return MARGINALIA, UnDeserializeableCategoryStringError
+}
 
+func (category Category) String() string {
 	return categoryStrings[category]
```
atmiguel: what does this do when `category` is out of bounds?

models/note.go
```diff
@@ -22,16 +23,33 @@ var categoryStrings = [...]string{
...
-		return "Unknown"
+var UnDeserializeableCategoryStringError = errors.New("String does not correspond to a Note Category")
+
+func DeserializeCategory(input string) (Category, error) {
+	for i := 0; i < len(categoryStrings); i++ {
+		if input == categoryStrings[i] {
+			return Category(i), nil
+		}
 	}
+	return MARGINALIA, UnDeserializeableCategoryStringError
```
atmiguel: Does go not let you return 0 here?

models/note.go
```diff
@@ -22,16 +23,33 @@ var categoryStrings = [...]string{
 	"predictions",
 }
 
-func (category Category) String() string {
-	if category < MARGINALIA || category > PREDICTIONS {
-		return "Unknown"
+var UnDeserializeableCategoryStringError = errors.New("String does not correspond to a Note Category")
+
+func DeserializeCategory(input string) (Category, error) {
+	for i := 0; i < len(categoryStrings); i++ {
```
atmiguel, line 29: Is it weird that we're doing a linear search here? Maybe I was wrong about the front-end not knowing about numbers we have corresponding to each category? Because something about this feels off

models/note.go
```diff
@@ -22,16 +23,33 @@ var categoryStrings = [...]string{
...
 type Note struct {
+	Id           int64     `json:"id"`
 	AuthorId     UserId    `json:"authorId"`
 	Content      string    `json:"content"`
 	CreationTime time.Time `json:"creationTime"`
 }
+
+func CreateNewNote(userId UserId, content string) *Note {
+	return &Note{
+		Id:           -1,
```
atmiguel: This is why it's weird to have `Id` on the note model. A note can exist before it has an Id.

services/noteservice/noteservice.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/atmiguel/cerealnotes/databaseutil"
+	"github.com/atmiguel/cerealnotes/models"
+)
+
+var NoteIdNotSet error = errors.New("The NoteId was not set")
+
+func StoreNewNote(
+	note *models.Note,
+) error {
```
atmiguel: this should maybe return the noteId

services/noteservice/noteservice.go
atmiguel: Why's that?

services/noteservice/noteservice.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if note.Id < 0 {
+		return NoteIdNotSet
+	}
+
+	return nil
+}
+
+func StoreNoteCategoryRelationship(
+	note *models.Note,
```
atmiguel: this function should just take the noteId, if there's no need for any of the other fields in Note

atmiguel (changes requested):

databaseutil/databaseutil.go
zingales: To clarify, you mean you shouldn't wonder if the `authorId` argument is of type `int64` or of type `UserId`? 

Honestly this file only gets waaay worse the more calls to the database we make. If you look at my later branches this file gets really out of hand. Personally i think we should consider adding an ORM package. Something that translates objects to sql and back. I think it would make our code look alot cleaner. But i also think that's a discussion we should make after we're done with the other CRs and we have all the pieces in front of us.

databaseutil/databaseutil.go
zingales: You're right, I'll change it.  I just makes me feel safer returning an invalid value on errors.

databaseutil/databaseutil.go
zingales: That should result in a database error, since an insert would make the note_id no longer unique In later branches i create a update note category relationship. That takes handles the existing note_id case: https://github.com/atmiguel/cerealnotes/blob/7a1a911bced216a0c74ae6bbe87f316970b12381/databaseutil/databaseutil.go#L181 If you want I we can only have a insertOrUpdate command, instead of two separate commands.

handlers/handlers.go
zingales: What if we call it `InstantiateNewNote`? Its my OOP-ness that makes me always want to create initializes, since its more modular. However, since its only used once here. I don't mind inlining it.

models/note.go
zingales: In the ui it'll be important to know the note id, so we can update and delete properly. The only time we won't know the id is when its new and hasn't been saved.

handlers/handlers.go
zingales: Its empty string as you can see in line 277. To continue down your idea, saving a new note will potentially be two api calls? insert note and insert category? That's do-able.

handlers/handlers.go
zingales: This is the only time we sanitize the string. Since the input is from the outside world this is the correct place to sanitize.

models/note.go
zingales: When you try to convert it to a type `Category` it will be either a cast or a compiler error.

models/note.go
zingales: IT DOESN'T AND I HATE IT!

models/note.go
zingales: is it weird that we are searching? or is it weird that the search is linear? Because if you want the front-end to be storage agnostic (which makes sense to me). Then you'll end up having to do a search. Though if you want we can store them in fancy objects to make it smaller than linear search but that seems like an over optimization at this point.

databaseutil/databaseutil.go
atmiguel: The only ORM i've ever used in that way is Ruby on Rails, so maybe i can't judge that pattern fairly. But I think the issue with this file isn't the fact that it has a lot of sql code. The issue is I think my idea was that this would be some sort of idealistic pure replacement of creating a builtin postgres function, which i was trying to avoid. I think that dream is dead, because it's only making me confused when i look at these function definitions and wonder why they've lost their type specificity

databaseutil/databaseutil.go
atmiguel: how are you defining invalid value?

databaseutil/databaseutil.go
atmiguel: Well my thought was that the front-end would either do a POST or a PUT. If they're doing a POST and the relationship already exists, I expect we'd return a 4xx. Conversely, if they're doing a PUT and the relationship does not exist, I expect we'd return a 4xx as well. If we just had an `insertOrUpdate` function we wouldn't be able to handle those cases

handlers/handlers.go
atmiguel: Yeah I was imagining two distinct api calls. It's the type of thing you squish together when you're looking to improve performance, but i see no reason to be optimizing for performance before there's an issue

handlers/handlers.go
atmiguel: cool yeah, let's inline it for now and we can talk about creating a "constructor" if we end up writing this once or twice more

handlers/handlers.go
atmiguel: Yeah it's from the outside world, but we set the rules of our api, not the outside world. This isn't "sanitizing", it's widening the amount of inputs we accept. The outside world should match exactly the api we set forth, otherwise they get a 4xx. I expect this won't be a problem, since we've standardized on these strings and made them lowercase all over the codebase

models/note.go
atmiguel: oh, wait. are you saying the out of bounds error is impossible?

atmiguel: like there's no way to pass an invalid `category` value into this function?

models/note.go
atmiguel: i'm not worried about the performance. it just feels like we're doing something wrong. this is just making me think that the front-end should maybe get like a map of ids to categories and then we never have to worry about converting from string -> category or vice versa anymore

databaseutil/databaseutil.go
```diff
@@ -94,6 +94,59 @@ func GetPasswordForUserWithEmailAddress(emailAddress string) ([]byte, error) {
...
+		}
+	}
+
+	if noteId == 0 {
+		return 0, QueryResultContainedNoRowsError
+	}
+
+	if err := rows.Err(); err != nil {
+		return 0, convertPostgresError(err)
+	}
```
atmiguel, line 127: okay new thought. I think what's tedious about this file is all the fucking boilerplate error handling shit. We should be able to call a function like `dbutil.QueryForASingleRow` that handles all this error shit internally and returns those same errors.

handlers/handlers.go
```diff
@@ -221,21 +222,22 @@ func HandleNoteApiRequest(
 ) {
 	switch request.Method {
 	case http.MethodGet:
-		note1 := &models.Note{
+
+		var noteMap noteservice.NoteMap = make(map[models.NoteId]*models.Note, 2)
```
atmiguel: for maps i typically go with the convention of `valuesByKey`, or in this case `notesById`

services/noteservice/noteservice.go
```diff
@@ -0,0 +1,48 @@
...
+func StoreNewNoteCategoryRelationship(
+	noteId models.NoteId,
+	category models.Category,
+) error {
+	if err := databaseutil.InsertNoteCategoryRelationship(int64(noteId), category.String()); err != nil {
+		return err
+	}
+
+	return nil
+}
```
atmiguel, line 35: I'm surprised you refrained from mentioning your hatred of the pointlessness of this file

services/noteservice/noteservice.go
```diff
@@ -0,0 +1,48 @@
...
+
+type NoteMap map[models.NoteId]*models.Note
+
+func (noteMap NoteMap) ToJson() ([]byte, error) {
+	// json doesn't support int indexed maps
+	datas := make(map[string]models.Note, len(noteMap))
```
atmiguel: i would say `notesByIdString` or `jsonableNotesById` instead of `datas`

services/noteservice/noteservice.go
```diff
@@ -0,0 +1,48 @@
...
+
+	for id, note := range noteMap {
+		datas[fmt.Sprint(id)] = *note
```
atmiguel: this is likely another reason people hate passing maps through http, because json is so limited. Another option here (that i expect you may dislike), is that in postgres, we could have the concept of a public and a private id. The current id would become the private id (meaning it would mostly stay in the database layer), and we'd assign a UUID as the public id for each note, upon creation. Then in our codebase, noteId would cease to be an integer, and would be a string instead. Thoughts?

atmiguel (changes requested):

databaseutil/databaseutil.go
zingales: it depends on the application and not everything has an invalid value (which is the crux of the issue). But for sql id a negative value is obviously invalid. But I changed it to zero, because that's convention.

databaseutil/databaseutil.go
zingales: Okay. Right now if you put an id that already exist it'll return as a 500 error. I'll add to technical debt to update `convertPostgresError` to handle duplicate key error.

models/note.go
zingales: That's fair. Let's postpone this conversation until we are looking at the pull request for the front end. Since for the backend this is a pretty simple fix.

databaseutil/databaseutil.go
zingales: I agree, but let's make a separate pull request to handle the databaseutil madness. I think its a bit out of scope for this pull request.

services/noteservice/noteservice.go
zingales: hahaha.

services/noteservice/noteservice.go
zingales: Both are valid options. I personally don't think the benefits of creating a uuid column is worth the change at least not right now.

models/note.go
zingales: I was wrong. I just tested it with this code and you can cast any number to Category. So i've added the check which returns "Unknown"

models/note.go
ethomas2: This took me a bit to figure out. I've been bitten by stuff like this before.

Firstly, the thing @atmiguel suggested actually does work. The following code compiles.
```go
type Category int

const (
	MARGINALIA Category = iota
	META
	QUESTIONS
	PREDICTIONS
)

var categoryStrings = []string{"A", "B", "C", "D"}

func DeserializeCategory(input string) (Category, error) {
	for i := 0; i < len(categoryStrings); i++ {
		if input == categoryStrings[i] {
			return Category(i), nil
		}
	}
	return 0, errors.New("some error")
}
```



That being said, figuring out type conversions in go can be tricky. I suspect you tried to do something innocent like this.
```go
func foo() Category {
  a := 0
  return a
}
```

^^^ that won't compile. Even though this will

```go
func foo() Category {
  return 0
}
```
This took me a bit to figure out, but I _think_ the reason the first `foo()` doesn't compile is because the compiler assigns type `int` to `a`. `int`and `Category` are **named types**. You can't assign a **named type** to another **named type**, even if they have the same **underlying type**. 

A named type is any type declared with the keyword `type`, along with any of the special **predeclared types** (`int`, `int64`, `float`, `string`, `bool`). You can think of the **predeclared types** as all being defined with the keyword `type` at the top of the file.

In contrast, this works fine
```go
type A []string

func bar() A {
  a := []string{}
  return a
}
```

^^^ `A` is a named type (because I declared it with `type`), `a` is an **unnamed type**. It's valid to assign an **unnamed type** to a **named type** as long as they have the same **underlying type**. 

Hopefully that explanation isn't too confusing. I think it'll probably be useful for you to read some of this stuff yourself
 [named type/unnamed type](https://stackoverflow.com/questions/19334542/why-can-i-type-alias-functions-and-use-them-without-casting/19334952#19334952)
[predeclared type](https://stackoverflow.com/questions/32983546/named-and-unnamed-types)
[underlying type](https://golang.org/ref/spec#Types)

also play around with some of it on your own

handlers/handlers.go
```diff
@@ -246,11 +249,83 @@ func HandleNoteApiRequest(
...
+			CreationTime: time.Now().UTC(),
+		}
+
+		noteId, err := noteservice.StoreNewNote(note)
+		if err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		responseWriter.Write([]byte(`{NoteId: "` + strconv.FormatInt(int64(noteId), 10) + `"}`))
```
atmiguel: is there a reason we can't just return the `noteId` in the content?

handlers/handlers.go
```diff
@@ -246,11 +249,83 @@ func HandleNoteApiRequest(
...
+		responseWriter.WriteHeader(http.StatusCreated)
+
 	default:
-		respondWithMethodNotAllowed(responseWriter, http.MethodGet)
+		respondWithMethodNotAllowed(responseWriter, http.MethodGet, http.MethodPost)
 	}
 }
 
+func HandleCategoryApiRequest(
```
atmiguel: should specify `NoteCategory`

handlers/handlers.go
```diff
@@ -246,11 +249,83 @@ func HandleNoteApiRequest(
...
+) {
+	switch request.Method {
+	case http.MethodPost:
+
+		type CategoryForm struct {
+			NoteId   int64  `json:"noteId"`
+			Category string `json:"category"`
+		}
+
+		noteForm := new(CategoryForm)
```
atmiguel: copy paste typo. Shouldn't be `noteForm`

handlers/handlers.go
```diff
@@ -246,11 +249,83 @@ func HandleNoteApiRequest(
...
+
+		if err := json.NewDecoder(request.Body).Decode(noteForm); err != nil {
+			http.Error(responseWriter, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		category, err := models.DeserializeCategory(strings.ToLower(noteForm.Category))
```
atmiguel: we shouldn't have to call `ToLower()`

handlers/handlers.go
```diff
@@ -284,8 +359,9 @@ func AuthenticateOrReturnUnauthorized(
 	authenticatedHandlerFunc AuthenticatedRequestHandlerType,
 ) http.HandlerFunc {
 	return func(responseWriter http.ResponseWriter, request *http.Request) {
+
 		if userId, err := getUserIdFromJwtToken(request); err != nil {
-			responseWriter.Header().Set("WWW-Authenticate", "Please log in to see this page")
+			responseWriter.Header().Set("WWW-Authenticate", `Bearer realm="`+request.URL.Path+`"`)
```
atmiguel, line 378: what's this?

models/note.go
```diff
@@ -22,7 +23,19 @@ var categoryStrings = [...]string{
 	"predictions",
 }
 
+var CannotDeserializeCategoryStringError = errors.New("String does not correspond to a Note Category")
+
+func DeserializeCategory(input string) (Category, error) {
+	for i := 0; i < len(categoryStrings); i++ {
+		if input == categoryStrings[i] {
+			return Category(i), nil
```
atmiguel, line 31: We'll just remember to think about this more later

paths/paths.go
```diff
@@ -8,7 +8,8 @@ const (
 	HomePage          = "/home"
 	NotesPage         = "/notes"
 
-	UserApi    = "/api/user"
-	SessionApi = "/api/session"
-	NoteApi    = "/api/note"
+	UserApi     = "/api/user"
+	SessionApi  = "/api/session"
+	NoteApi     = "/api/note"
+	CategoryApi = "/api/Category"
```
atmiguel: the url should be all lowercase

atmiguel: and should include note in it somewhere

services/noteservice/noteservice.go
```diff
@@ -0,0 +1,48 @@
...
+	category models.Category,
+) error {
+	if err := databaseutil.InsertNoteCategoryRelationship(int64(noteId), category.String()); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type NoteMap map[models.NoteId]*models.Note
```
atmiguel: `NotesById`?

handlers/handlers.go
```diff
@@ -246,11 +249,83 @@ func HandleNoteApiRequest(
...
 	}
 }
 
+func HandleCategoryApiRequest(
+	responseWriter http.ResponseWriter,
+	request *http.Request,
+	userId models.UserId,
+) {
+	switch request.Method {
+	case http.MethodPost:
```
atmiguel: we should come to some agreement on this. The main issue with it being a post is that i would expect it to return some id. And since we're specifying an id in the request this feels much more like a PUT. Though that means the `noteId` needs to move into the url. Whether in the query params or as part of the url I don't have a strong preference. We should just be consistent in whatever we choose.

atmiguel (changes requested):

handlers/handlers.go
zingales: according to the docs the header is supposed to look like this `WWW-Authenticate: <type> realm=<realm>` type bearer means its a token/cookie. realm is supposed to describe what you were trying to access. I though the path would do a good enough trick. 

docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

handlers/handlers.go
zingales: I'm not sure what you mean by that. The response to this is a json object with the `noteId` of the newly published note. 

are you asking why i had to format the noteid into a string?

handlers/handlers.go
zingales: You want it to read `HandleNoteCateogryApiRequest`?

handlers/handlers.go
zingales: I'm okay with removing the distinction between post and put for note category. I agree right now the distinction doesn't make sense. So this handler will eventually have a `get` a `put` and a `delete` but no `post` ya?

models/note.go
zingales: in regards to what? i can add a comment if you'd like

handlers/handlers.go
atmiguel: I'm saying just the `noteId` is a valid json object. We could just return that. If you're not a fan of that we can just return something like `{"id": 3}`. The key should be camelCase and the value shouldn't be in quotes. Additionally, the way I'd expect us to return that is by json serializing some struct

handlers/handlers.go
atmiguel: yeah that seems reasonable. Just saying `Category` seems a bit confusing

models/note.go
atmiguel: sorry, i'm referring to an earlier conversation we had about considering whether or not to just pass around int values for this enum to the front end. We can just deal with it later.

handlers/handlers.go
atmiguel: sounds right

handlers/handlers.go
```diff
@@ -246,11 +249,85 @@ func HandleNoteApiRequest(
...
 
+func HandleNoteCateogryApiRequest(
+	responseWriter http.ResponseWriter,
+	request *http.Request,
+	userId models.UserId,
+) {
+	switch request.Method {
+	case http.MethodPut:
+
+		id, err := strconv.ParseInt(request.URL.Query().Get("id"), 10, 64)
```
atmiguel, line 309: We should verify that `id` is the one and only query param. And verify that it is an integer. If either of those things aren't true, we should return a 400. Right now, for some of those things, we're returning a 500

models/note.go
```diff
@@ -22,7 +23,19 @@ var categoryStrings = [...]string{
...
+		}
+	}
+	return MARGINALIA, CannotDeserializeCategoryStringError
```
atmiguel: We should just return `0` here. I'd also be okay with returning `-1` instead of the usual `0`, because `0` is actually valid here

atmiguel: actually, ignore that `-1` comment. There are always gonna be times where the falsy value is also valid. But this should just be `0` hard-coded

services/noteservice/noteservice.go
```diff
@@ -0,0 +1,48 @@
...
+	if err := databaseutil.InsertNoteCategoryRelationship(int64(noteId), category.String()); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type NotesById map[models.NoteId]*models.Note
+
+func (noteMap NotesById) ToJson() ([]byte, error) {
```
atmiguel: the variable name should be `notesById` as well

atmiguel (changes requested):

handlers/handlers.go
zingales: I'll add this to our Technical backlog. Because i think eventually creating a method like respondWithMethodNotAllowed is the right call.

zingales: https://trello.com/c/Oz9QbNVj/57-add-call-to-make-sure-there-are-not-extra-query-paramters

atmiguel: I wouldn't respond with MethodNotAllowed, but other than that, I agree

atmiguel (approved):

PR #34, opened 2018-09-13
Delete note api

zingales:


static/js/base.js
```diff
@@ -0,0 +1,32 @@
...
+        dataType: type,
+        data: data,
+        success: callback
+      });
+    };
+  });
+
+  jQuery.prototype.getDOM = function() {
+    if (this.length === 1) {
+        return this[0];
```
zingales, line 23: Interestingly `[0]` is faster than `.get(0)` https://learn.jquery.com/using-jquery-core/faq/how-do-i-pull-a-native-dom-element-from-a-jquery-object/

zingales:
in deference to https://github.com/atmiguel/cerealnotes/pull/43

PR #38, opened 2018-09-16
Mock Db test framework Option A

zingales:


integration_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/atmiguel/cerealnotes/models"
+	"github.com/atmiguel/cerealnotes/paths"
+	"github.com/atmiguel/cerealnotes/routers"
+)
+
+func TestLoginOrSignUpPage(t *testing.T) {
+	mockDb := &DiyMockDataStore{}
+	env := &handlers.Environment{mockDb}
+
+	handlers.SetEnvironment(env)
```
ethomas2, line 24: Not fan of this global environment. I realize you're following the same patter that you had with databaseutil. 

For one thing, if you ever set your tests to run in parallel (which they aren't right now) setting this global env isn't threadsafe. Test1 might overwrite Test2's env in the middle of test 2 running. Even worse, two tests might try to write to write to the global env at the same time and clobber each other's writes. You'll get some garbage struct for your env

To make your tests run in parallel, add `t.Parallel()` to the top of every test
https://golang.org/pkg/testing/#T.Parallel

To have `go test` detect race conditions like these, run it with `go test -race`

integration_test.go
```diff
@@ -0,0 +1,129 @@
...
+	handlers.SetTokenSigningKey([]byte(""))
+
+	server := httptest.NewServer(routers.DefineRoutes())
+	defer server.Close()
+
+	resp, err := http.Get(server.URL)
```
ethomas2, line 30: It's good practice to always `defer resp.Body.Close()` after making an http request
https://stackoverflow.com/questions/23928983/defer-body-close-after-receiving-response

```
If the returned error is nil, the Response will contain a non-nil Body which
the user is expected to close. If the Body is not closed, the Client's
underlying RoundTripper (typically Transport) may not be able to re-use a
persistent TCP connection to the server for a subsequent "keep-alive" request.
```

ethomas2 (commented):
Test stuff looks good overall to me. Minor comments

ethomas2:
btw, there's a library out there which will take an interface and auto generate a mock struct that implements that interface

https://github.com/vektra/mockery

integration_test.go
zingales: Even if i don't use the respone.body?

integration_test.go
zingales: I'm open to suggestions. I just don't know how to create non-end to end tests without mocking a db. And i'm not sure how to do that without A) having a global variable or B) passing in the environment object to all function calls. Which just gets annoying.

zingales: I guess i could make all the handlers functions, methods on the environment struct, but i'm not a fan of that.

zingales:
closed in favor of https://github.com/atmiguel/cerealnotes/pull/39

PR #39, opened 2018-09-17
Mock Db test framework option B

zingales:


handlers/tokenutil_test.go
```diff
@@ -0,0 +1,64 @@
...
+	bob, err := handlers.CreateTokenAsString(env, num, 1)
+	if err != nil {
+		panic(err)
+	}
+
+	token, err := handlers.ParseTokenFromString(env, bob)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(bob)
```
atmiguel: why are we printing in our tests?

handlers/tokenutil_test.go
```diff
@@ -0,0 +1,64 @@
...
+			t.FailNow()
+		}
+		fmt.Printf("%v %v", claims.UserId, claims.StandardClaims.ExpiresAt)
+	} else {
+		fmt.Println("Token claims could not be read")
+		t.FailNow()
+	}
+}
+
+func assert(tb testing.TB, condition bool, msg string, v ...interface{}) {
```
atmiguel: what's `tb`?

handlers/tokenutil_test.go
```diff
@@ -0,0 +1,64 @@
...
+	if !condition {
+		_, file, line, _ := runtime.Caller(1)
+		fmt.Printf("\033[31m%s:%d: "+msg+"\033[39m\n\n", append([]interface{}{filepath.Base(file), line}, v...)...)
```
atmiguel: what's going on here?

handlers/tokenutil_test.go
```diff
@@ -0,0 +1,64 @@
...
+// ok fails the test if an err is not nil.
+func ok(tb testing.TB, err error) {
+	if err != nil {
+		_, file, line, _ := runtime.Caller(1)
+		fmt.Printf("\033[31m%s:%d: unexpected error: %s\033[39m\n\n", filepath.Base(file), line, err.Error())
+		tb.FailNow()
+	}
+}
+
+// equals fails the test if exp is not equal to act.
```
atmiguel: this comment would be unnecessary if the variables just had the full names

integration_test.go
```diff
@@ -0,0 +1,231 @@
...
+	mockDb := &DiyMockDataStore{}
+	env := &handlers.Environment{mockDb, []byte("")}
+
+	server := httptest.NewServer(routers.DefineRoutes(env))
+	defer server.Close()
+
+	// Create testing client
+	client := &http.Client{}
+	{
+		// jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
```
atmiguel: do we still need this comment?

integration_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Content-Type", contentType)
+	return client.Do(req)
+}
+
+func printBody(resp *http.Response) {
```
atmiguel, line 160: this function is never called

models/category.go
```diff
@@ -0,0 +1,62 @@
+package models
+
+import (
+	"errors"
+)
+
+type Category int
```
atmiguel: the filename and everything should probably be noteCategory, instead of just category

migrations/tools/createDatabases.sql
```diff
@@ -0,0 +1,2 @@
+CREATE DATABASE cerealnotes_test;
```
atmiguel, line 1: why do we need a separate test db?

models/datastore_test.go
```diff
@@ -0,0 +1,141 @@
...
+func ClearAllValuesInTable(db *models.DB) {
+	for _, val := range tables {
+		if err := ClearValuesInTable(db, val); err != nil {
+			panic(err)
+		}
+	}
+}
+
+func ClearValuesInTable(db *models.DB, table string) error {
+	// db.Query() doesn't allow varaibles to replace columns or table names.
```
atmiguel: typo in `variables`

models/datastore_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+func ClearAllValuesInTable(db *models.DB) {
+	for _, val := range tables {
+		if err := ClearValuesInTable(db, val); err != nil {
+			panic(err)
+		}
+	}
+}
+
+func ClearValuesInTable(db *models.DB, table string) error {
```
atmiguel, line 37: why do we need to clear the values in the tables?

models/datastore_test.go
```diff
@@ -0,0 +1,141 @@
...
+func ok(tb testing.TB, err error) {
+	if err != nil {
+		_, file, line, _ := runtime.Caller(1)
+		fmt.Printf("\033[31m%s:%d: unexpected error: %s\033[39m\n\n", filepath.Base(file), line, err.Error())
+		tb.FailNow()
+	}
+}
+
+// equals fails the test if exp is not equal to act.
+func equals(tb testing.TB, exp, act interface{}) {
```
atmiguel: if we're using these functions everywhere we should put them in a test_util file

atmiguel (changes requested):
I can't say I agree with putting the database logic into our models. What was the main reason behind that decision?

handlers/tokenutil_test.go
zingales: I don't know. As you can see the TB is what the package calls it.

handlers/tokenutil_test.go
zingales: this fancyness gets the file and line number of the calling method.

handlers/tokenutil_test.go
zingales: i'll remove the comment

integration_test.go
zingales: ya but its very useful for debugging. When debugging i use it to print, but then remove the print body once the test is successful so not to clutter up the stdout

migrations/tools/createDatabases.sql
zingales: This test run against the database, we don't want any bad code to leave extra unexpected data in our database. So for isolation i created a seperate DB.

models/datastore_test.go
zingales: to make sure every run of the tests runs on the same input.

integration_test.go
zingales: no

models/datastore_test.go
zingales: okay we'll just have to be careful about circular dependencies.

PR #41, opened 2018-09-18
Propagate error messages to front end

ethomas2:


static/js/login_or_signup.js
```diff
@@ -132,34 +132,34 @@ $(function() {
     };
 
     attachSubmitClickHandler(signupFormMetadata, (formDataAsJsonString) => {
-        $.post('/api/user', formDataAsJsonString, (responseBody, _, request) => {
-            if (request.status === 201) {
+        $.post('/api/user', formDataAsJsonString, (responseBody, _, jqXHR) => {
```
zingales: Adrian and I have decided to use long winded variable names, so what's the long winded version of `jqXHR`. Reading online i suspect going by our conventions it should be `$XmlRequest` or `$XmlHttpRequest` The `$` signifies that its a jQuery object.

static/js/login_or_signup.js
```diff
@@ -132,34 +132,34 @@ $(function() {
...
                 alert('Unknown successful status');
             }
-        }).fail((request) => {
-            if (request.status === 409) {
+        }).fail((jqXHR) => {
+            if (jqXHR.status === 409) {
                 alert('Email address already in use');
             } else {
-                alert('Unknown error');
+                alert('Unknown error ' + jqXHR.responseText);
```
zingales: Nitpick: Can you change this to Unexpected error, since your printing signifies that it is "known"

zingales (changes requested):

static/js/login_or_signup.js
ethomas2: I think `xhr` is one of those rare cases where the shorter name is actually more readable than the long name. Just because `xhr` is such a common thing in js and very quickly recognizable as "oh, it's one of _those_ objects". I expressed similar thoughts [here](https://github.com/atmiguel/cerealnotes/pull/16#discussion_r178424547). Basically, I prefer not to deviate from the commonly used names for things, even if that name isn't a name I would've chosen on my own.

Also fwiw, I got the name from the jquery docs. 
https://api.jquery.com/jquery.get/#jQuery-get-url-data-success-dataType

That being said, I like the idea of prepending jquery objects with `$`, so how do you feel about `$xhr`? If you feel strongly that it should be expanded then I prefer  `$XmlHttpRequest`, but it's a bit unfortunate because the whole reason I changed the name is because conceptually this isn't a request, it's a response. If I change it to `$XmlHttpRequest` then i'm just putting request back in the name. Not a huge deal because at least it's clear that it's an `xhr` object, it just bugs me.

zingales: what about `$XmlHttpResponse`? Also i think Adrian's previous answer to your thoughts is still valid. For what its worth,  I had absolutely no idea what `jqxhr` was and had to look it up.

zingales: Also i think Adrian's previous answer to your thoughts is still valid. For what its worth,  I had absolutely no idea what `jqxhr` was and had to look it up.

ethomas2: eh, i'll just use `$XmlHttpRequest`. I don't think there's such a thing as `XmlHttpResponse` (at least i've never seen it). I think the response is an attribute of an xhr.

Also, I probably wouldn't've realized what `jqxhr` was except that I saw it in the jquery docs. Since I was on a jquery doc page, I realized the `jq` meant jquery and `xhr` meant `xhr`. But if I saw something called `$xhr`, I would totally recognize that that's an `xhr` object.

zingales (approved):

PR #49, opened 2018-09-22
Dockerfile and docker-compose.yml changes

ethomas2:
* Switch to multi-stage dockerfile
* Remove postgres dockerfile
* Remove Godeps (Godeps is deprecated)
* Add dependencies to Gopkg.toml
* Add vendor


docker-compose.yml
```diff
@@ -1,30 +1,24 @@
...
-    volumes:
-      - ./docker_db/pgdata:/pgdata
-    ports:
-      - "5432"
-    environment:
-      - POSTGRES_USER=docker
-      - POSTGRES_PASSWORD=docker
-      - PGDATA=/pgdata
+    db:
+        image: postgres:9.6
```
zingales: nice! i like not needed an extra dockerfile!

docker-compose.yml
```diff
@@ -1,30 +1,24 @@
...
-      - TOKEN_SIGNING_KEY=AllYourBase
-    build:
-      context: ./
-      args:
-        app_env: ${APP_ENV} 
-    volumes:
-      - .:/go/src/github.com/atmiguel/cerealnotes/
-    ports:
-      - "8080:8080"
-    links:
```
zingales, line 30: I think having the links/depends on is important. It makes sure docker starts the db image before the server image.

zingales (changes requested):
As we talked about offline. I don't want vendor checked in.

docker-compose.yml
ethomas2: Done

zingales (approved):

